PR #209, opened 2022-01-09
Adds a minimal support for user phrases.

### LINKED ISSUE #6: 支援使用者詞彙
more description is needed

zonble:
The user phrases are stored in a text file in the same format of "data.txt" contained in the application bundle. A user can use shift and arrow keys like left and right to select a range of the text in the composing buffer, and use enter key to add the text into the text file. Once a user does so, the input method reloads the entire text file into a copy of `FastLM`.

Now, creating an instance `BlockReadingBuilder` requires two language models, one is the original language model, while another one is for user phrases. When `BlockReadingBuilder::build()` is called, the builder gets unigrams from both language model.

Since the user phrases are in a text file, a user can edit it with any text editor. I did not let the input method app to watch the change of text file since I don't think user may often edit it with external text editors. I just add a menu item to let user to reload user phrases manually.

I did not block duplicate entries of user phrases. Since the text file is editable, a user can just remove the duplicated entries anytime and reload the user phrases.

Fixes #6, #85 and #131 .


ShikiSuen:
@zonble 請問，您目前這套系統交給使用者自訂語彙時，使用者填每一行的格式時的頻率是不是也要填那種經過換算過後的指數？

我記得每個字詞的頻率換算時用到的 norm 數值是要在遍歷所有詞的總頻率之後才能算出來的。

ShikiSuen:
我在這邊分享一下我的設想好了，打算接下來這周我這邊先開工。
https://github.com/ShikiSuen/vChewing-macOS/issues/12

這個設計圖紙當中沒有考慮到小麥注音的ㄅ半傳統輸入模式的情況，但可以再追加。
也就是在編譯的時候只篩留單個字的結果而已。

lukhnos:
LGTM.

PR #210, opened 2022-01-09
Fixes a bug on building the vector for unigrams from both global language model and user phrases.

zonble:
There was a bug. If there is a custom user phrase, it override the phrases in the original language model.


lukhnos:
LGTM.

PR #211, opened 2022-01-10
Fixes the bug that I forgot to create the user phrases folder.

zonble:
There was a legacy user override model which creates a folder and a
plist file. If a user uses McBopomofo for years, the folder would
exist. However, when the old override model was removed, I forgot
to create the folder for the new user phrase file.

The bug would let the users with new installation of McBopomofo unable
to add user phrases.


lukhnos:
LGTM.

PR #212, opened 2022-01-10
Converts most of the Objective-C classes into Swift

zonble:
I guess it is time to modernize the code base.

The following classes are now in Swift

- AppDelegate
- InputSourceHelper
- NonModalAlertWindowController
- PreferencesWindowController

The candidate UI becomes a local Swift Package.

It is how the project looks like after the pull request

Screen Shot 2022-01-11 at 2 40 55 AM


zonble:
BTW, I use NSURLSession to do version checking instead if NSURLConnection.

lukhnos (approved):
LGTM.

> BTW, I use NSURLSession to do version checking instead if NSURLConnection.

I didn't realize McBopomofo was still using `NSURLConnection`. Indeed a much needed modernization.

PR #216, opened 2022-01-11
Introduces custom excluded phrases

zonble:
The pull request uses a FastLM to load the phrases that that a user want to exclude from the phrase database. Now the BlockReadingBuilder reads data from three language models, the main input method language model, the user phrases, and excluded phrase. The builder looks up for Unigrams from the first and second models and use the last one as a filter.

A user can start to edit the file storing phrases to exclude from the input menu and then click on the "reload user phrases" menu item to make new excluded phrases to take effect. Actually I think it is an advanced function and there might be a Wiki page for it.

This solves requests like issue #123.

### DIFF
--- a/Source/Engine/Gramambular/BlockReadingBuilder.h
+++ b/Source/Engine/Gramambular/BlockReadingBuilder.h
@@ -38,7 +38,7 @@ namespace Formosa {
         
         class BlockReadingBuilder {
         public:
-            BlockReadingBuilder(LanguageModel *inLM, LanguageModel *inUserPhraseLM);
+            BlockReadingBuilder(LanguageModel *inLM, LanguageModel *inUserPhraseLM, LanguageModel *inExcludedPhrasesLM);
             void clear();
             
             size_t length() const;
@@ -58,6 +58,8 @@ namespace Formosa {
             vector<string> readingsAtRange(size_t begin, size_t end) const;
 
             Grid& grid();
+
+            bool checkIfUnigramExistInVector(Unigram& unigram, vector<Unigram>vector);
                         
         protected:
             void build();
@@ -73,13 +75,17 @@ namespace Formosa {
             
             Grid m_grid;
             LanguageModel *m_LM;
-            LanguageModel *m_UserPhraseLM;
+            LanguageModel *m_userPhraseLM;
+            LanguageModel *m_excludedPhrasesLM;
             string m_joinSeparator;
         };
         
-        inline BlockReadingBuilder::BlockReadingBuilder(LanguageModel *inLM, LanguageModel *inUserPhraseLM)
+        inline BlockReadingBuilder::BlockReadingBuilder(LanguageModel *inLM,
+                                                        LanguageModel *inUserPhraseLM,
+                                                        LanguageModel *inExcludedPhrasesLM)
             : m_LM(inLM)
-            , m_UserPhraseLM(inUserPhraseLM)
+            , m_userPhraseLM(inUserPhraseLM)
+            , m_excludedPhrasesLM(inExcludedPhrasesLM)
             , m_cursorIndex(0)
             , m_markerCursorIndex(SIZE_MAX)
         {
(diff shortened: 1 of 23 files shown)


zonble:
By the way, I moved the instance of these language models to another file. Since the main input controller is already a little bit too big, and the models are actually global but should not be a part of the input controller, I created a new "Language Model Manager" as the facade for all of the language models.

zonble:
Temporarily close the PR since it is better to hide the additional language models to a new abstract language model.

Source/Engine/Gramambular/BlockReadingBuilder.h
```diff
@@ -58,6 +58,8 @@ namespace Formosa {
             vector<string> readingsAtRange(size_t begin, size_t end) const;
 
             Grid& grid();
+
+            bool checkIfUnigramExistInVector(Unigram& unigram, vector<Unigram>vector);
```
lukhnos: No longer used. Remove?

Source/Engine/Gramambular/BlockReadingBuilder.h
```diff
@@ -197,7 +197,17 @@ namespace Formosa {
         {
             return m_grid;
         }
-        
+
+        inline bool BlockReadingBuilder::checkIfUnigramExistInVector(Unigram& unigram, vector<Unigram>vector)
```
lukhnos: No longer used.

Source/Engine/McBopomofoLM.cpp
```diff
@@ -0,0 +1,92 @@
...
+{
+    m_languageModel.close();
+    m_languageModel.open(languageModelDataPath);
+}
+
+void McBopomofoLM::loadUserPhrases(const char* userPhrasesDataPath,
+                                   const char* excludedPhrasesDataPath)
+{
+    m_userPhrases.close();
+    m_userPhrases.open(userPhrasesDataPath);
```
lukhnos: `LanguageModelManager` passes `""` to `userPhrasesDataPath` when creating Plain Bopomofo's LM. Perhaps allow `nullptr` here, and only call `open` if `userPhrasesDataPath` is not `nullptr`?

Same for `excludedPhrasesDataPath`.

I realize calling `::open("", O_RDONLY)` in `FastLM` will likely lead to a safely-guarded failure, but one more guard shouldn't hurt…

Source/Engine/McBopomofoLM.h
```diff
@@ -0,0 +1,31 @@
...
+                         const char* m_excludedPhrasesDataPath);
+
+    const vector<Bigram> bigramsForKeys(const string& preceedingKey, const string& key);
+    const vector<Unigram> unigramsForKey(const string& key);
+    bool hasUnigramsForKey(const string& key);
+
+protected:
+    FastLM m_languageModel;
+    FastLM m_userPhrases;
+    FastLM m_excluddePhrases;
```
lukhnos: Nit: `m_excludedPhrases`?

Source/LanguageModelManager.mm
```diff
@@ -0,0 +1,155 @@
...
++ (void)loadDataModels
+{
+    LTLoadLanguageModelFile(@"data", gLanguageModelMcBopomofo);
+    LTLoadLanguageModelFile(@"data-plain-bpmf", gLanguageModelPlainBopomofo);
+}
+
++ (void)loadUserPhrasesModel
+{
+    gLanguageModelMcBopomofo.loadUserPhrases([[self userPhrasesDataPathMcBopomofo] UTF8String], [[self excludedPhrasesDataPathMcBopomofo] UTF8String]);
+    gLanguageModelPlainBopomofo.loadUserPhrases("", [[self excludedPhrasesDataPathPlainBopomofo] UTF8String]);
```
lukhnos: Let `loadUserPhrases` accept `nullptr`, and pass `nullptr` here?

lukhnos (approved):
LGTM.

PR #217, opened 2022-01-12
Fix regression in vertical candidate UI

lukhnos:
The table view style must be set before it's added as the scroll view's
content view. See [1].

[1] https://github.com/openvanilla/McBopomofo/blob/1.1/Source/CandidateUI/VTVerticalCandidateController.m#L110


zonble:
Done.

PR #219, opened 2022-01-13
Various new functions and UI components

zonble:
- Half-width punctuation support
- Yahoo KeyKey! style tooltip UI
- Yahoo KeyKey! style notifier UI
- Emacs style function keys to move the cursor and the candidate list


lukhnos (approved):
LGTM.

PR #220, opened 2022-01-14
Use a more tolerant parser for user phrases

lukhnos:
A generic key-value blob reader, KeyValueBlobReader, is implemented to
allow more flexibility in user-editable files. For example, this allows
comments in the file, as well as tolerating leading or trailing spaces,
tabs, or even Windows CR LF line endings.

Unit tests are supplied for KeyValueBlobReader although they are not
part of the Xcode project. A separate CMakeLists.txt is provided.

UserPhrasesLM is refactored to use KeyValueBlobReader. A small stylistic
change is appiled to reduce "using namespace" uses, but otherwise no
major style changes were applied to UserPhrasesLM.

Please note that McBopomofo's user phrase LM uses the value in a
key-value pair as the reading, and the key as the actual "value". We
don't plan to change that order so that we don't have to migrate data.

std::string_view is used to allow efficient reference to char buffers
and interop with std::string (and so no c_str() is needed). C++17 is now
enabled for the project to enable the use of std::string_view.

Copyright headers are added to McBopomofoLM and UserPhrasesLM.

### DIFF
--- a/McBopomofo.xcodeproj/project.pbxproj
+++ b/McBopomofo.xcodeproj/project.pbxproj
@@ -32,6 +32,7 @@
 		6ACA41FD15FC1D9000935EF6 /* MainMenu.xib in Resources */ = {isa = PBXBuildFile; fileRef = 6ACA41F015FC1D9000935EF6 /* MainMenu.xib */; };
 		6ACA41FF15FC1D9000935EF6 /* main.m in Sources */ = {isa = PBXBuildFile; fileRef = 6ACA41F415FC1D9000935EF6 /* main.m */; };
 		6ACA420215FC1E5200935EF6 /* McBopomofo.app in Resources */ = {isa = PBXBuildFile; fileRef = 6A0D4EA215FC0D2D00ABF4B3 /* McBopomofo.app */; };
+		6ACC3D3F27914F2400F1B140 /* KeyValueBlobReader.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 6ACC3D3E27914F2400F1B140 /* KeyValueBlobReader.cpp */; };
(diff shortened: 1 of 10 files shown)


zonble:
A question. The coding style looks quite Googlish. Are we changing our C++ coding style?

lukhnos:
> A question. The coding style looks quite Googlish. Are we changing our C++ coding style?

I think it's open for discussion. What do you think?

You're right that it's Googlish but not quite. For example `KeyValueBlobReader.cpp` would have to be called `key_value_blob_reader.cc` and enums would have to be like `State::kEnd` instead of `State::END` if I wanted to follow the style more closely. I guess I'm trying the water to see what a style shift would be like for McBopomofo—but granted we probably won't make wholesale changes, if we ever decided to switch style.

From my end, it's just that I use cpplint, and the unit tests are also written using Google Test. 2-space idents are better for 80-column lines that have a lot of `std::` uses.

zonble:
We have use Webkit style for years I just want to know if we will change the style of the existing code, and the style that I should follow when I am doing C++ portion.

BTW, adopting C++19 makes the code looks quite elegant.

zonble:
I just saw that your commits became "unverified". I did not know what went wrong....

zonble (approved):

lukhnos:
> I just saw that your commits became "unverified". I did not know what went wrong....

I forgot to sign them. I'll force-push the signed commits…

> We have use Webkit style for years I just want to know if we will change the style of the existing code, and the style that I should follow when I am doing C++ portion.

That's a very good point. Let me treat this PR as a deviation from our established practice. Given the contributors' bandwidth, coding style is not really one more thing that one needs to think about.

PR #223, opened 2022-01-14
Extracts preferences and emacs key detection from the input controller

zonble:
The input controller is already too big so I extract the tasks of management of preference keys and detect emacs style keys from it.

I added a new Swift class, Preference. It uses property wrappers to manage the user defaults in a modern way. It also presents the input controller from handling the default values and let the rules for all of the default values in one place.

Anther class EmacsKeyHelper helps to make the process to map emacs like function keys more expressive.

Also added icon to the keyboard layouts in the preference window.


zonble:
Finally, I bring the legacy VXHanConvert back. I don't know the exact difference of its conversion result with those by OpenCC, but I guess some users might miss the legacy converter.

It should be the last thing that I want to add for McBopomofo 2.0.

ShikiSuen:
@zonble OpenCC is the current authority of Chinese conversion, especially how it handles the conversion to detailed variants of Traditional Chinese used in HongKong and Taiwan. It even supports the conversion between Chinese and simplified ideographs used in Japan.

VXHanConvert does a good job in recovering 锺 馀 谘 when converting to Simplified Chinese. However, in Simplified Chinese only the 锺 gets recovered (as family name used by 锺兴民、锺南山, etc.).

P.S.: I clearly remembered that VXHanConvert generated 馀 谘 made me scold fiercely several times on Weibo by random individuals. During that time, the VXHanConvert built in Yahoo KeyKey was the only choice of typing Simplified Chinese without sacrificing the choice of a combo Daqian bopomofo IME.

P.P.S: The difference between the VXHanConvert and OpenCC are huge. At least, VXHanConvert does not support phrase-based conversion. OpenCC at least can convert "著→着" correctly in phrases (not less than 2 kanji ideographs), while VXHanConvert is not capable of doing this.

ShikiSuen:
@zonble (continue)

In short: You want an effective conversion to Simplified Chinese (in lieu of an authentic Simplified Chinese language model data file). Regarding the converted Simplified Chinese results, VXHanConvert's results are far less effective comparing to OpenCC's results. Simplified Chinese is not your born language, though you can still master it by relearning the 现代汉语词典　just like what I did with 世一辭典 when I was learning Traditional Chinese by myself.

ShikiSuen:
BTW. The current way how OpenCC gets applied in McBopomofo is still not ideal enough.
Let me introduce @lotem 's Squirrel IME: https://github.com/rime/squirrel

If I'm not wrong ( Lotem please point it out if I'm wrong), when you type something enough to call a list of candidates, Squirrel first ask OpenCC to convert the candidate list to the target language mode. After that, there might be duplicated entries (e.g. 幹干 ㄍㄢˋ to Simplified Chinese 干): Squirrel has a special unit called "simplifier" which automatically filters out those duplicated entries except the one which has the highest candidate priority. After finishing this step, the processed candidate list gets sent to the candidate controller.

zonble:
@ShikiSuen  I guess you can file a ticket to OpenCC and ask them to improve the Chinese conversion engine, so it can fixed the issue for all projects using the library. Maintaining a Chinese conversion engine is out of my personal scope.

ShikiSuen:
@zonble I am not asking you to fix that. Sorry for letting you confused.
If I'm not wrong, have you encountered issues with OpenCC?
You can tell me what issue is happening, and I'll forward it to @Byvoid's repository.

P.S.: I just noticed that you added both VXHanConvert and OpenCC. My appreciations.

ghost:
佛振的 OpenCC 模式恐需要小麥注音更換為康熙繁體詞庫，不然很難解決樓上所講的「著→着」的轉換問題：
如果詞庫是台灣繁體中文的話，單個漢字的轉換表內本來就沒有「着」，而 OpenCC 轉單字搭配 Simplifier 必然會導致「著」與「着」少一樣，這樣就會干擾對單個漢字的輸入。
在不換詞庫的前提下，這個小缺陷要想解決的話，倒也不難，就是在確保 BPMFBase 有「着」的情況下、在 Phrase.occ 當中給予「着」僅次於「著」的權重。

ShikiSuen:
@AidenPearceDev 喔也是。我差點忘記 Squirrel 的台灣繁體是用康熙繁體藉由 OpenCC 轉換過去的。
雖然這轉換也是用力過猛……凶兇不分就是了，相關的轉換模組非常缺乏考究。

這個話題可以單獨扯出來討論。先讓這 PR 就這樣。

lukhnos (approved):
LGTM.

ShikiSuen:
@zonble Typo:

image

Source/InputMethodController.mm
```diff
@@ -388,6 +302,14 @@ - (void)setValue:(id)value forTag:(long)tag client:(id)sender
 
 #pragma mark - IMKServerInput protocol methods
 
+- (NSString *)_convertToSimplifiedChinese:(NSString *)text
+{
+    if (Preferences.chineneConversionEngine == 1) {
+        return [VXHanConvert convertToSimplifiedFrom:text];
+    }
+    return [OpenCCBridge convertToSimplified:text];
```
ShikiSuen, line 310: Lack of "else". I dunno whether an 'else' is necessary in this context in Swift.

Source/PreferencesWindowController.swift
```diff
@@ -106,37 +125,37 @@ private let kDefaultKeys = "123456789"
...
+            Preferences.candidateKeys = nil
             NSSound.beep()
             return
         }
 
         selectionKeyComboBox.stringValue = keys
-        if keys == kDefaultKeys {
-            UserDefaults.standard.removeObject(forKey: kCandidateKeys)
+        if keys == Preferences.defaultKeys {
+            Preferences.candidateKeys = nil
```
ShikiSuen, line 156: 這裡好像不能這麼寫，得把 nil 改成 Preferences.defaultKeys。
不然的話，一旦選了非預設的選字鍵，再想選回來就選不回來了，因為永遠都是 nil。

Source/PreferencesWindowController.swift
```diff
@@ -106,37 +125,37 @@ private let kDefaultKeys = "123456789"
...
         }
     }
 
     @IBAction func changeSelectionKeyAction(_ sender: Any) {
         let keys = (sender as AnyObject).stringValue.trimmingCharacters(in: CharacterSet.whitespacesAndNewlines)
         if keys.count != 9 || !keys.canBeConverted(to: .ascii) {
-            selectionKeyComboBox.stringValue = kDefaultKeys
-            UserDefaults.standard.removeObject(forKey: kCandidateKeys)
+            selectionKeyComboBox.stringValue = Preferences.defaultKeys
+            Preferences.candidateKeys = nil
```
ShikiSuen, line 149: 這裡好像也不能這麼寫，得把 nil 改成 Preferences.defaultKeys。
不然的話，一旦選了錯誤的選字鍵，只有 comboBox 的內容正常了，但 candidateKeys 裡面的東西是空的。
我這邊測試發現，一旦這方面出錯，刪掉 preferences plist 還不算，得重新開機。 cc @zonble

ShikiSuen:
@zonble 這條回覆是我的猜測。
我上文提到的兩處與 nil 有關的錯誤的真正原因，恐怕與下面這條指令有關：
```
        var candidateSelectionKeys = Preferences.candidateKeys ?? Preferences.defaultKeys
```
不然就是`Preferences.candidateKeys`一刻也不能 nil。

但我目前也只能先按照我上面那兩處修改來保守治療了。

Source/PreferencesWindowController.swift
ShikiSuen: This one already gets fixed in https://github.com/openvanilla/McBopomofo/pull/232/commits/9a3f5806fbf640022f32f783f903e6692a1d20ff

PR #237, opened 2022-01-16
Allows users to type Latin letters while using shift + letter keys.

zonble:
Fixes issue #162.

I just treat the letters as punctuation symbols and put them to the punctuation table. When a user type shift + a, the input controller find the unigrams using the key "_letter_A" from the language model.


ShikiSuen:
Is this dynamically compatible with non-standard alpha-numeral keyboards like dvorak?

lukhnos (approved):
LGTM

PR #241, opened 2022-01-18
Unit tests and bug fixes

zonble:
1. Adds test cases for Swift based components like candidate windows, preferences, selection key validators, etc.
2. Reformats Swift code.
3. Makes the tooltip window to stay in the screen.
4. Notifies the users that it is not suggested to add phrases by shift left/right when phrase replacement mode, or model based Chinese conversion is on.
5. Fixes typos.


zonble:
@lukhnos , I would like to ask you about this line of code:

https://github.com/openvanilla/McBopomofo/blob/master/Packages/CandidateUI/Sources/CandidateUI/VerticalCandidateController.swift#L312

Actually, it is possible that firstVisibleRow might be larger than selectedRow. So, `UInt(selectedRow - firstVisibleRow)`  may raise an exception since it might be Unit(-1). Now I just the method to return if firstVisibleRow is larger than selectedRow. I do not know if what I do is right.

Thank you for your time.

McBopomofoTests/PreferencesTests.swift
```diff
@@ -144,9 +144,9 @@ class PreferencesTests: XCTestCase {
     }
 
     func testChineneConversionEngine() {
```
ShikiSuen: func test**Chinese**ConversionEngine() {

Consider batch replace "hinene" to "hinese" among all files in the repo? You can stage the changes later in your git software client.

Source/InputMethodController.mm
```diff
@@ -1456,6 +1456,14 @@ - (void)_showCurrentMarkedTextTooltipWithClient:(id)client
...
+    else if (Preferences.phraseReplacementEnabled) {
+        NSString *message = NSLocalizedString(@"Phrase replacement mode is on. Not suggested to add phrase in the mode.", @"");
+        [self _showTooltip:message client:client];
+    }
+    else if (Preferences.chineseConversionStyle == 1 && Preferences.chineseConversionEnabled) {
+        NSString *message = NSLocalizedString(@"Model based Chinese conversion is on. Not suggested to add phrase in the mode.", @"");
+        [self _showTooltip:message client:client];
+    }
     else if (length == 1) {
         NSString *messsage = [NSString stringWithFormat:NSLocalizedString(@"You are now selecting \"%@\". You can add a phrase with two or more characters.", @""), text];
```
ShikiSuen: It's "message", not "mes**s**sage". Consider performing batch-replacement fix?

Source/InputMethodController.mm
```diff
@@ -55,6 +55,8 @@
 using namespace OpenVanilla;
 
 static const NSInteger kMinKeyLabelSize = 10;
+static const NSInteger kMinMarkRangeLength = 2;
+static const NSInteger kMaxMarkRangeLength = 6;
```
ShikiSuen, line 59: 要是能從 blockReadingBuilder.h 直接讀取 MaximumBuildSpanLength 的話？

lukhnos:
> @lukhnos , I would like to ask you about this line of code: https://github.com/openvanilla/McBopomofo/blob/master/Packages/CandidateUI/Sources/CandidateUI/VerticalCandidateController.swift#L312
> 
> Actually, it is possible that firstVisibleRow might be larger than selectedRow. So, `UInt(selectedRow - firstVisibleRow)` may raise an exception since it might be Unit(-1). Now I just the method to return if firstVisibleRow is larger than selectedRow. I do not know if what I do is right.

I see. I had never considered this possibilty…

Let's be defensive then? That is:

```swift
if (selectedRow >= firstVisibleRow) {
  keyLabelStripView.highlightedIndex = UInt(selectedRow - firstVisibleRow)
} else {
  // boundary condititon, set no label in the strip view as highlighted
  keyLabelStripView.highlightedIndex = UInt.max
}
```

lukhnos (approved):
LGTM.

PR #244, opened 2022-01-18
Writes templates the first time custom data is used

lukhnos:
Now that we allow comments in the custom data files, this change writes
localized templates as well as basic instructions. Links to McBopomofo
User's Manual are also provided.


zonble:
LGTM.

PR #246, opened 2022-01-19
Make UserPhrasesLM more tolerant

lukhnos:
This lets UserPhrasesLM consumes as much user data as possible before
bailing. This makes it more tolerant to data errors and will not fail
entirely just because the user has one faulty line in a data file.

Also removes FastFM from the benchmarking suite.

This also runs the CMake-based C++ tests as part of the GitHub CI.


Source/Engine/UserPhrasesLMTest.cpp
```diff
@@ -0,0 +1,57 @@
...
+    fprintf(f, "bar \n"); // error line
+    fprintf(f, "argh baz\n");
+    int r = fclose(f);
+    ASSERT_EQ(r, 0);
+
+    UserPhrasesLM lm;
+    lm.open(tmp_name.c_str());
+    ASSERT_TRUE(lm.hasUnigramsForKey("foo"));
+    ASSERT_FALSE(lm.hasUnigramsForKey("bar"));
+    ASSERT_FALSE(lm.hasUnigramsForKey("baz"));
```
zonble: I am confused. It looks like the line "argh baz\n" is valid, why do we assert false here?

zonble (approved):
I think it is OK, but there is a test case makes me confused. :)

Source/Engine/UserPhrasesLMTest.cpp
lukhnos: > I am confused. It looks like the line "argh baz\n" is valid, why do we assert false here?

I updated the tests to clarify how the errors are handled—when the parser encounters an error, it will stop parsing, and so the third line, even if valid, won't be parsed. I've also updated the test keys and values to make the intent clear.

In addition, I realized that `PhraseReplacementMap` also needed to be more tolerant, and I've made the changes and added the corresponding test.

PR #254, opened 2022-01-21
Converts main.m and the installer into Swift

zonble:
Also added the Credits.rtf file to show the links to the user manual and FAQ in the about window.


Source/Installer/BundleTranslocate.m
```diff
@@ -0,0 +1,30 @@
+#import "BundleTranslocate.h"
```
lukhnos, line 1: Use the same copyright header from `BundleTranslocate.h`?

Source/main.swift
```diff
@@ -0,0 +1,109 @@
+//
```
lukhnos: Let's use the new copyright header such as the one in `AppDelegate.swift` in this PR.

lukhnos (approved):
LGTM. Thanks!

ShikiSuen:
NotoSansCJK is not shipped in macOS by default. What about changing the RTF font to systemfont or "Helvetica Neue" or "SF Pro Display (install separately but macOS can parse it to system font)" in the macOS TextEdit app if possible?

PR #259, opened 2022-01-28
Applies a state centric design to McBopomofo

zonble:
MaBopomofo has a fat input controller, and the push request tears the input controller to three parts: key handler, input states and state handler, and arranges them in a clear one-way data flow inspired by architectures like Redux:

```
Event -> Key Handler -> State -> UI or output
```

Currently the input controller has following issues:

*  The variables for detecting current state are stored in many places. For example, we use if the candidate window is visible to identify if the user is choosing candidates, the index of the cursor is stored in the grammar builder, the composing buffer is a member variable of the controller... it is hard to trace what the input controller is doing.
* The variables mentioned above could be accessed anytime and it will eventually cause bugs. For example, the controller can still ask to open the candidate window and modify the composing buffer even it is deactivated. Lots of these variables are only useful in a specific scenario and they should be managed.
* Since the input states are not managed, we are also hardly to add new states, and functions needs new states, in the future, such as associated phrases in plain bopomofo mode. 

The PR separates the key input processing and UI/text output with states. The key handler does not access the input client directly but it yields states. For example, when a user sends a delete key, the key handler may find that all contents in the input buffer are erased so it yields an "empty" state to replace the existing "inputting" state. Another example is, if a user sends an enter key, it yields "committing" state and then "empty" state.

State objects are immutable and they have members only for the current state. For example, the composing buffer and the cursor index are stored in the "inputting" state, so they will not be accessed when the input controller is inactive. The candidates are stored only in the "choosing candidate" state as well.

The definitions of the states and the key handler are extracted from the the input controller. The input controller now only update UI and input buffer by the states, and the methods for UI and output are collected in one place.

It is how the code looks like

``` objc
    KeyHandlerInput *input = [[KeyHandlerInput alloc] initWithEvent:event isVerticalMode:useVerticalMode];
    BOOL result = [_keyHandler handleInput:input state:_state stateCallback:^(InputState *state) {
        [self handleState:state client:client];
    }           candidateSelectionCallback:^{
        NSLog(@"candidate window updated.");
    }                        errorCallback:^{
        NSBeep();
    }];

    return result;
```

The key handler yields the states in to the `stateCallback`, and the controller pass the states to the state handler. We can easily to monitor the change of the state if we add a log here.

The PR does not complete everything that I want to do. Which candidate is highlighted is still stored in the candidate window controller. It should be moved to the "choosing candidate" state object but it needs to refactor the architecture to present candidates. The cursor index is still in the grammar builder and the BPMF reading is still in the key handler, but they should be moved to the "inputting" state object.


zonble:
Another benefit is, now we can do unit-tests to the key handler. We can mock input and do assertion on the states returned from the state block.

lukhnos (approved):
LGTM. Thanks for doing this!

PR #261, opened 2022-01-29
Converts the input controller to Swift

zonble:
Since the C++ engine is move to the key handler class, now we can convert the input controller to Swift.

The test cases for the key handler are converted to Swift as well. The input mode becomes a Swift-friendly typed enum.

Some minor modification:

- Give a minimal cell width to the horizontal candidate window. It makes the window size for the punctuation list to be fixed. Currently it always changes the size when it goes to another page and it makes me dizzy.
- Lazy load the language models, so the McBopomofo user does not need to load the model for Plain Bopomofo.


Source/InputMethodController.swift
```diff
@@ -0,0 +1,547 @@
...
+
+    override func deactivateServer(_ client: Any!) {
+        keyHandler.clear()
+        self.handle(state: .Empty(), client: client)
+        self.handle(state: .Deactivated(), client: client)
+    }
+
+    override func setValue(_ value: Any!, forTag tag: Int, client: Any!) {
+        let newInputMode = InputMode(rawValue: value as? String ?? InputMode.bopomofo.rawValue)
+        LanguageModelManager.loadDataModel(newInputMode)
```
lukhnos, line 130: Could you check if `setValue` is called only once per instantiation of the input method controller? Or could this be called every time we activate/deactivate the controller?

Since `LanguageModelManager.loadDataModel()` doesn't check if the data model is already loaded, I just wanted to make sure that we are not re-loading data models repeatedly.

Source/LanguageModelManager.h
```diff
@@ -22,14 +22,14 @@
...
 #import <Foundation/Foundation.h>
-#import "UserOverrideModel.h"
-#import "McBopomofoLM.h"
+#import "KeyHandler.h"
 
 NS_ASSUME_NONNULL_BEGIN
 
 @interface LanguageModelManager : NSObject
 
 + (void)loadDataModels;
```
lukhnos: Now that this is a test-only method, perhaps rename it to call out that it's only used by tests (in `KeyHandlerBopomofoTests` that is)?

lukhnos (approved):
Just one question regarding the lazy loading semantics and one comment on a method name, otherwise LGTM!

Source/LanguageModelManager.h
zonble: OK, let me do this.

Source/InputMethodController.swift
zonble: OK. I did additional checks to prevent loading data models repeatedly.

PR #263, opened 2022-01-30
Implements the associated phrases function.

zonble:
Since we use states manage the input flow in McBopomofo, implementing this function becomes easy. What I did is to create a new state, Associated Phrases state, and let the key handler to emit such a state just after emitting a Committing state.

When the input method controller is under Associated Phrase state, it shows the candidate window with a tooltip, and only accept candidate keys with the shift key. The key handler uses the characters without modifiers in an NSEvent object to find if there is any matching candidate label, so I added a new member "inputTextIgnoringModifiers" to KeyHandlerInput.

I use KeyValueBlobReader to read the associated phrases. I use the cin file from OpenVanilla project but removed the head and tail of the file to pass KeyValueBlobReader's validation.


zonble:
I am not quite sure about how to handle enter key under the associated phrases mode, since I am not a plain Bopomofo user. Currently I just allow users to use shift + enter to choose a phrase.

lukhnos (approved):
LGTM. Two comments:

First, not sure how much you want to go down the rabbit hole, but in OV we show Shift+1 labels (so `⇧1`, `⇧2`, etc.) in the Associate Phrases' candidate panel to indicate that shift key. Or at least should we use the tooltip to prompt the user to use Shift+1~9?

And:

> I am not quite sure about how to handle enter key under the associated phrases mode, since I am not a plain Bopomofo user. Currently I just allow users to use shift + enter to choose a phrase.

I don't think we need anything for Enter combinations. We rely on Shift+[1-9] only, and anything else (except for the candidate paging keys) should cancel the associated phrases state.

zonble:
I added the ⇧ symbol on the candidate list. :)

PR #264, opened 2022-01-30
Loads the user phrases just after the files are changes.

zonble:
The revision uses a FSEventStream to monitor the change of the folder that stores user phrases.


lukhnos (approved):
LGTM

PR #268, opened 2022-02-01
Various bug fixes for the inputting and marking states

zonble:
The PR contains following fixes

1. Fix the bug that the app may crash when marking a range with Emoji consists of two or more UTF16 code points
2. Disallow user to use shift + arrow keys to add new phrases when the string length and readings count do not match
3. Show a tooltip when the cursor is in a phase whose length and readings count do not match

There were two issues that we have to deal with when we are using shift and arrow keys to mark a range in the composing buffer. 

The first one is, the exact index of a character and its position in an NSString may differ and we need to do conversion to make the results correct. The composing buffer is an NSAttributedString and we use NSRange to mark the range in it, and its length is just the count of UTF16 code points, but there might be some emoji consists of two or more UTF16 code points. When there are such Emoji symbols, we should use their lengths to jump to another position when using shift and arrow keys to make NSString happy but not simply add or subtract one.

So, now we can set marking ranges in a string with such symbols.

Screen Shot 2022-02-01 at 8 24 54 AM

Another issue is about the string length and readings count. McBopomofo allows users to input two or more readings to compose a character, or three and more readings for a phrase with two characters, for example, ㄙㄨㄛˇ-ㄧˇ can compose ∴. When this happens, we cannot know what are the exact range of the readings when we select a range in the string, and I think we should not let users to add new phrases in this case.

On the other hand, users are also hardly to find where the cursor is. So I added a tooltip when the cursor is in the phrase whose string length and readings count do not match. A screenshot is attached:

Screen Shot 2022-02-02 at 1 03 04 AM

Additional revisions

- Allow users to input punctuations with control key and keys like ",", ".".
- Prevent the key handler to see keys like page up/down as BPMF keys.
- Remove unused localization string.


Source/Installer/en.lproj/Localizable.strings
```diff
@@ -37,3 +37,6 @@
 "Warning" = "Warning";
 "Input method may not be fully enabled. Please enable it through System Preferences > Keyboard > Input Sources." = "Input method may not be fully enabled. Please enable it through System Preferences > Keyboard > Input Sources.";
 "Continue" = "Continue";
+
+"Model based Chinese conversion is on. Not suggested to add phrase in the mode." = "Model based Chinese conversion is on. Not suggested to add phrase in the mode.";
+"Model based Chinese conversion is on. Not suggested to add phrase in the mode." = "Model based Chinese conversion is on. Not suggested to add phrase in the mode.";
```
lukhnos: Duplicate line here. Delete?

lukhnos (approved):
LGTM.

Source/Installer/en.lproj/Localizable.strings
zonble: OK. Let me do this later.

PR #269, opened 2022-02-02
Refactor the Bopomofo syllable library for better thread safety 

lukhnos:
The library has accumulated a lot of cruft. A more thorough refactoring requires further efforts, but this PR addresses some of the worst offenders:

- Global singletons should be initialized in a thread-safe manner. Since C++11, this is guaranteed with the scoped (local) static initialization.
- `std::` should be spelt out in full; `using namespace std;` should never appear in any modern C++ code.
- `uint16_t` is now used instead of `short`.
- The UTF-8 literal is used so that we don't have to do `#ifdef _MSC_VER` for MS C++ anymore (which fully supports the UTF-8 literal).
- `constexpr` is used where appropriate.
- Default copy and move constructors are used.

In addition,

- One wildcard use is removed. That was an overengineered design to allow using string identifiers to set keyboard layout.
- Simple UTF-8 heuristics are used to convert composed Bopomofo strings back to syllables.

As a result, all OpenVanilla dependencies are removed from this set of changes.

Lastly, I've clang-formatted the files. While I understand that we've been on WebKit style, technically Formosana is not part of McBopomofo :), and for a non-trival C++ library like this, the tooling story (CLion, cpplint, standard and up-to-date style gudie and ongoing discussions, number of people who know the style etc.) makes it so much easier to adopt the style.

Note that I didn't change any public method names, in order to keep the refactoring "blast radius" small. I didn't bother to rewrite some of the wordy iterator uses, nor did I go over every line to remove dead code.

Unit tests are added and now we have at least some test coverage.

The C++ code was first written back in 2006. It wouldn't pass code review at any reputable C++ house today (not that it would back in 2006). This gives it a much needed refresh.

### DIFF
--- a/Source/Engine/Mandarin/Mandarin.cpp
+++ b/Source/Engine/Mandarin/Mandarin.cpp
@@ -35,7 +35,8 @@ namespace Mandarin {
 
 class PinyinParseHelper {
  public:
-  static const bool ConsumePrefix(string& target, const string& prefix) {
+  static const bool ConsumePrefix(std::string& target,
+                                  const std::string& prefix) {
     if (target.length() < prefix.length()) {
       return false;
     }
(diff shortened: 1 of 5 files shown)


Source/Engine/Mandarin/Mandarin.cpp
```diff
@@ -23,991 +20,1236 @@
...
-{
-    string result;
-    #define APPEND(c) if (m_syllable & c) result += (*BopomofoCharacterMap::SharedInstance().componentToCharacter.find(m_syllable & c)).second
-    APPEND(ConsonantMask);
-    APPEND(MiddleVowelMask);
-    APPEND(VowelMask);
-    APPEND(ToneMarkerMask);
-    #undef APPEND
-    return result;
+#define ASSIGNKEY1(m, vec, k, val) \
```
zonble, line 996: I am a little confused. I do not understand why are we reusing the vector but not to create a new one? I am glad to know the benefits. 😀

zonble (approved):
I can clearly understand your code and I think it is OK to merge. I just have two question

1. Why do we reuse the vector in macros like ASSIGNKEY 1, 2, 3?
2. The naming of macros like ASSIGNKEY1 is a little confusing to me.

Source/Engine/Mandarin/Mandarin.cpp
lukhnos: That really has no benefits, and can really be written now with things like `m[k] = std::vector{val1, val2, ...}`. Even back in those days it could have been written with `m[k].push_back(...); m[k].push_back(...);`. It was really just an oversight.

`ASSIGNKEYx` could use a better macro name (or better, as an inline function), and it really means "assign the key at state _x_ to the Bopmofo syllable".

PR #272, opened 2022-02-02
Bug fixes

zonble:
I used wrong API to save user phrases.


McBopomofoTests/KeyHandlerBopomofoTests.swift
```diff
@@ -1,3 +1,26 @@
+// Copyright (c) 2011 and onwards The McBopomofo Authors.
```
lukhnos: 2022 for files created this year? Same for other files here.

lukhnos (approved):
LGTM.

McBopomofoTests/KeyHandlerBopomofoTests.swift
zonble: OK. Let me update it.

PR #273, opened 2022-02-02
Use backward-compatible Swift in certain places

lukhnos:
Older Swift compiler does not allow declaring certain variables that
have the same names with those outside of their scope, even though the
scoping rules should allow them. This makes the code buildable with
Xcode 12.4 again.


zonble (approved):
Thank you for the PR.

PR #276, opened 2022-02-03
Minor improvements

zonble:
* Adds a preference to move the cursor to next node after selecting a candidate in the MS IME style (candidate after cursor). Since it is very possible that a user uses the left key to go backward to select a candidate, the new preference can help users to jump back to where he or she is from.
* Notifies the users if they are selecting a duplicated phrase when using shift left/right.
* Adds more unit tests.


zonble:
A GitHub action file to do building and testing using Xcode 12 is also added.

lukhnos (approved):
LGTM

PR #282, opened 2022-02-13
Adds symbols and emoji characters

zonble:
- Updates the contents of the symbols
- Adds a hidden function to output Bopomofo readings that helps to edit the phrases


lukhnos (approved):
LGTM

PR #283, opened 2022-02-13
Bopomofo refactoring

lukhnos:
Just some small clean-up to pave the way for further refactoring.


zonble:
It looks safe to delete these functions. :)

PR #284, opened 2022-02-13
Remove Tests/

lukhnos:
Deleting the directory seems safe. Checked that no Python code is referring to these files. We can always revive them if still needed.


zonble:
LGTM.

PR #285, opened 2022-02-14
Minor bug fix and polish

lukhnos:
This makes the Half-Width Punctuation toggle work again, and copyedits some of the UI strings.


zonble:
LGTM.

zonble (approved):

PR #297, opened 2022-02-19
Remove the "using namespace" usage in McBopomofo's C++ and Objective-C++ code

lukhnos:
This brings the code in line with modern C++ practices, and paves the way for future refactoring.

This also reformats the Objective-C and Objective-C++ code in McBopomofo in WebKit-style. I didn't set the column width (which is also the WebKit style default) and I also didn't ask clang-format to reformat pointer alignments (`Foo *` vs `Foo*`), so as to minimize the diff.

I was actually surprised by how small the reformat diff is (most of the diff here is about `using namespace` removal).


zonble:
It looks OK. It looks like most of the modifications are by formatting tools.

zonble (approved):

PR #302, opened 2022-03-01
Add phrases

zonble:


lukhnos:
LGTM, though the diff is huge. I noticed that instead of sorting by key alone, it seems the data is also sorted by length (so 2-character pharses come first, then 3-character phrases in the second group).

PR #313, opened 2022-04-29
Updates the style of the preference window

zonble:
Currently we put all settings in the same view. The PR adds tabs in the preference window to put advanced settings to another tab.

Screen Shot 2022-04-29 at 10 59 13 AM

### DIFF
--- a/Source/PreferencesWindowController.swift
+++ b/Source/PreferencesWindowController.swift
@@ -24,6 +24,14 @@
 import Cocoa
 import Carbon
 
+fileprivate extension NSToolbarItem.Identifier {
+    static let basic = NSToolbarItem.Identifier(rawValue: "basic")
+    static let advanced = NSToolbarItem.Identifier(rawValue: "advanced")
+}
+
+fileprivate let kWindowTitleHeight: CGFloat = 78
+
+
 // Please note that the class should be exposed as "PreferencesWindowController"
 // in Objective-C in order to let IMK to see the same class name as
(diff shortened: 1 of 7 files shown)


Source/PreferencesWindowController.swift
```diff
@@ -155,3 +180,73 @@ import Carbon
...
+        windowRect.origin.y = window.frame.maxY - (viewFrame.height + kWindowTitleHeight)
+        window.setFrame(windowRect, display: true, animate: true)
+        window.contentView?.frame = view.bounds
+        window.contentView?.addSubview(view)
+    }
+
+    @objc func showBasicView(_ sender: Any?) {
+        use(view: basicSettingsView)
+        window?.toolbar?.selectedItemIdentifier = .basic
+        window?.title = NSLocalizedString("Basic", comment: "")
```
lukhnos, line 203: I was wondering if we should let the window title follow the tab title:

截圖 2022-04-28 下午8 43 08

Then I found that both Terminal and Xcode do this. Interestingly, Finder doesn't… (Its window title is always "Finder Preferences".)

It seems that even Apple apps aren't consistent here. My only argument for preferring Finder's treatment is that in Exposé it shows the window title, and you'll know that it's Finder's preferences, whereas for Terminal's and Xcode's it's not always clear to me that what "General" is about. But then again, I guess because of how we present the window, McBopomofo's preferences window doesn't even show up in Exposé.

So I feel this is fine. Just leaving a note about my observations.

lukhnos (approved):

PR #314, opened 2022-05-03
Adds a setting to let user assign custom user phrase folder

zonble:
The screenshots are attached.

Screen Shot 2022-05-04 at 12 25 53 AM

Screen Shot 2022-05-04 at 12 25 56 AM

### DIFF
--- a/Source/Preferences.swift
+++ b/Source/Preferences.swift
@@ -414,28 +414,39 @@ class Preferences: NSObject {
 }
 
 extension NSNotification.Name {
-    static var useCustomUserPhraseLocationDidChange = NSNotification.Name(rawValue: "PreferencesUseCustomUserPhraseLocationDidChangeNotification")
-    static var customUserPhraseLocationDidChange = NSNotification.Name(rawValue: "PreferencesCustomUserPhraseLocationNotification")
+    static var userPhraseLocationDidChange = NSNotification.Name(rawValue: "UserPhraseLocationDidChangeNotification")
 }
 
 extension Preferences {
 
+    static func postUserPhraseLocationNotification() {
+        let location: String = {
+            if !useCustomUserPhraseLocation {
+                return UserPhraseLocationHelper.defaultUserPhraseLocation
+            }
+            if customUserPhraseLocation.isEmpty {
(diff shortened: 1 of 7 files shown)


Source/Preferences.swift
```diff
@@ -401,3 +404,38 @@ class Preferences: NSObject {
...
+
+    @UserDefault(key: kUseCustomUserPhraseLocation, defaultValue: false)
+    @objc static var useCustomUserPhraseLocation: Bool {
+        didSet {
+            if useCustomUserPhraseLocation == false {
+                customUserPhraseLocation = nil
+            } else if customUserPhraseLocation == nil || customUserPhraseLocation == "" {
+                customUserPhraseLocation = UserPhraseLocationHelper.defaultUserPhraseLocation
+            }
+            NotificationCenter.default.post(name: .useCustomUserPhraseLocationDidChange, object: useCustomUserPhraseLocation)
```
lukhnos: If all the code paths will lead to `.customUserPhraseLocationDidChange` being posted, do we still need `.useCustomUserPhraseLocationDidChange`? Also only the formet gets observed in AppDelegate.

Source/PreferencesWindowController.swift
```diff
@@ -178,6 +185,68 @@ fileprivate let kWindowTitleHeight: CGFloat = 78
...
+    @IBAction func changeUserPhraseEnabledAction(_ sender: Any) {
+        guard let control = sender as? NSControl else {
+            return
+        }
+        let enabled = control.integerValue > 0
+        if enabled {
+            if Preferences.customUserPhraseLocation == nil || Preferences.customUserPhraseLocation == "" {
+                Preferences.customUserPhraseLocation = UserPhraseLocationHelper.defaultUserPhraseLocation
+            } else {
+                Preferences.customUserPhraseLocation = Preferences.customUserPhraseLocation
```
lukhnos: Just to confirm: This is setting the property to its current value. Is this to trigger the posting of  the `.customUserPhraseLocationDidChange` notification?

lukhnos (approved):
Was just wondering if we can just define one notification, otherwise LGTM!

zonble:
Now I use only one notification to notify app delegate to reload data when changing the location.

PR #315, opened 2022-05-06
Modifies the behavior of sole tone mark

zonble:
McBopomofo allows users to input sole tone mark like "ˇ", "ˋ", "ˊ", and "˙ " into the composing buffer when lots of other Zhuyin input methods decide it is an error to input a tone mark with any syllable. It helps users to type tone marks when they want to edit custom phrases but this behavior is annoying.

- When a user wants to type ㄚ(8 in the standard layout), he or she may wrongly type 7(˙) and then the tone mark is just sent to the buffer.
- When a user wants to type 以(u3)， he or she may type 3u since the keys requires two hands and it causes the output become ˇㄧ.

The PR changes the behavior. Now we have error tolerance and allow the users to type ˇㄧ to become 以, but ˇㄅ is still an error input. If a user really want to type a sole tone mark, now a space key after the tone mark is required.

### DIFF
--- a/McBopomofoTests/KeyHandlerBopomofoTests.swift
+++ b/McBopomofoTests/KeyHandlerBopomofoTests.swift
@@ -493,10 +493,7 @@ class KeyHandlerBopomofoTests: XCTestCase {
             } errorCallback: {
             }
         }
-        XCTAssertTrue(state is InputState.Inputting, "\(state)")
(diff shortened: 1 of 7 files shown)


zonble:
This is also implemented in the fcitx version. PR 40 and 41.

lukhnos:
LGTM.

PR #316, opened 2022-05-06
Implements the feature to backup phrases using git

zonble:
The PR adds a hook that runs git commands when a user adds a new user phrase. The function could be enabled or disabled from the preference.

Screen Shot 2022-05-06 at 12 24 23 PM


lukhnos:
Would you like to carry over the approach you use in fcitx5-mcbopomofo to run a script for the tasks instead?

zonble:
Done. :)

lukhnos (approved):
LGTM

PR #323, opened 2022-05-31
Fixes the flow to fix the nodes at front of the composing buffer. 

zonble:
Fixes the flow to fix the nodes at front of the composing buffer.

In the previous PR I put the code before user override models suggestion. It causes the fixes node to be set to another score. Fixed.

I used a workaround yesterday to fix the nodes at front by code like "fixNode(offset + 1)", but it was buggy, and sometimes causes crashes.

Once we change the order to walk through the tree, we cannot just reset the score of the nodes crossing or ending at a given location, but the nodes in the range of the spanning length.

In a range of 台灣人 may contain nodes like 台灣 and 人, once we only reset the nodes crossing or ends at the end index 3, it won't effect the node 台灣, but if we handle only the index, it won't apply to 人. We have to change the way to find effect nodes to reset both 台灣 and 人. Merely using (index + 1) looks right but it is actually wrong.

Also let the tab key handler to handle empty states correctly.

### DIFF
--- a/Source/Engine/Gramambular/Grid.h
+++ b/Source/Engine/Gramambular/Grid.h
@@ -52,6 +52,7 @@ class Grid {
   size_t width() const;
   std::vector<NodeAnchor> nodesAt(size_t location);
   std::vector<NodeAnchor> nodesCrossingOrEndingAt(size_t location);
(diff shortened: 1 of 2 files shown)


lukhnos:
LGTM

PR #327, opened 2022-07-09
Use Gramambular2

### LINKED ISSUE #309: Consider allowing very long composing buffer
Many other input methods do not limit composing buffer length and therefore do not need to evict overflow segments.

Right now McBopomofo limits the buffer length due to the exponential runtime in the `walk()` algorithm, but since the simple model has a very small "blast radius" in terms of how a new segment would affect previous or next segments, we may consider only "walk" the -10/+10 readings from the cursor. Some audit of the algorthim (which was implemented long time ago) may also be needed.

We would still want to impose a max length though, beyond which we won't allow user input any more text. 200 readings for example, which is almost the length of a small article already.

lukhnos:
這個 PR 跟 fcitx5-mcbopomofo 的 [PR#58](https://github.com/openvanilla/fcitx5-mcbopomofo/pull/58) 類似，導入 gramambular2，對 KeyHandler 做必要修改。

- Keep Engine code in sync with fcitx5-mcbopomofo
- Keep associtaed phrases support to McBopomofoLM
- Update KeyHandler and InputState to use Gramambular2
- Remove the composing buffer size restriction
- CI settings update to run Gramambular2 tests

Fixes #309.


lukhnos:
Added commits to clean up UI (as well as two localized strings that are no longer used) and to support the (reading, value) candidate pairs.

Screen Shot 2022-07-11 at 10 19 33 PM

zonble (approved):
OK. I tried this version and it works exactly as our Linux version. :)

PR #335, opened 2022-08-08
Format Objective-C++ code

lukhnos:
A custom .clang-format is introduced to minimize diff. It is evident
that the Objective-C++ code has been pretty much following the WebKit
style throughout the years.


zonble:
LGTM

PR #344, opened 2022-09-15
Bump deployment target to macOS 10.13

lukhnos:
Xcode 14 has removed support for deployment targets below macOS 10.13. This change allows the project to be built with the most recent Xcode versions.

This also migrates the usage of one deprecated NSWorkplace method.


zonble:
LGTM.

PR #347, opened 2022-10-18
Stop calling setMarkedText upon (de)activation

### LINKED ISSUE #346: Certain IMK calls time out upon (de)activitation in some apps
We've received user reports that McBopomofo got killed macOS in certain macOS version + app + architecture combinations. Inspecting the diagnostic reports further revealed that certain IMK remote calls timed out in those cases. 

t's really more like a problem on the app's side, since it's they that ulmately decide how those IMK calls are handled once the calls get across the process boundary. This being said, we can mitigate the problem by avoiding making those calls in those situations.

lukhnos:
Fixes #346

Certain apps don't handle -setMarkedText:selectedRange:replacementRange: consistently upon IMK session (de)activation. It's an effectively no-op anyway, and so this simplifies the (de)activation call.

Also updates the state machine transitions so that entering the deactivated state will cause the state machine to transit to the empty state. This is to guarantee that the state machine is always in the empty state upon IMK session activation.

### DIFF
--- a/Source/InputMethodController.swift
+++ b/Source/InputMethodController.swift
@@ -121,14 +121,13 @@ class McBopomofoInputMethodController: IMKInputController {
 
         keyHandler.clear()
         keyHandler.syncWithPreferences()
-        self.handle(state: .Empty(), client: client)
+
         (NSApp.delegate as? AppDelegate)?.checkForUpdate()
     }
 
     override func deactivateServer(_ client: Any!) {
         currentClient = nil
         keyHandler.clear()
-        self.handle(state: .Empty(), client: client)
         self.handle(state: .Deactivated(), client: client)
     }
 
@@ -289,6 +288,9 @@ extension McBopomofoInputMethodController {
 
         if let newState = newState as? InputState.Deactivated {
             handle(state: newState, previous: previous, client: client)
+            // Force transition to Empty, so that when activateServer is
+            // invoked again, the controller is already in the Empty state.
+            state = .Empty()
         } else if let newState = newState as? InputState.Empty {
             handle(state: newState, previous: previous, client: client)
         } else if let newState = newState as? InputState.EmptyIgnoringPreviousState {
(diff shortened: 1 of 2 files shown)


Source/InputMethodController.swift
```diff
@@ -121,14 +121,13 @@ class McBopomofoInputMethodController: IMKInputController {
 
         keyHandler.clear()
         keyHandler.syncWithPreferences()
-        self.handle(state: .Empty(), client: client)
```
zonble, line 124: Wondering why this is removed?

Source/InputMethodController.swift
```diff
@@ -121,14 +121,13 @@ class McBopomofoInputMethodController: IMKInputController {
...
+
         (NSApp.delegate as? AppDelegate)?.checkForUpdate()
     }
 
     override func deactivateServer(_ client: Any!) {
         currentClient = nil
         keyHandler.clear()
-        self.handle(state: .Empty(), client: client)
```
zonble, line 131: Also wondering why this is removed?

Source/InputMethodController.swift
lukhnos: When activateServer gets called, there's no need to enter the Empty state handler—after this PR, the state machine is guranteed to be in the Empty state (previously it could be in the Deactivated state).

Plus, log analysis showed that some apps may not be able to handle `setMarkedText` remote calls consistently, and the Empty state handler does need to call that.

Source/InputMethodController.swift
lukhnos: Similar to `activateServer`. When this gets called, the text input client is already closing a session, and so it's not necessary to call `setMarkedText`, which not all applications handle consistently in this state. Plus, this PR made the change such that we make the state machine enter the Empty state after entering the Deactivated state, and so this Empty state handling is no longer needed.

zonble:
LGTM :)

PR #355, opened 2023-01-29
Implements a Big5 code input mode

zonble:
I know some traditional Bopomofo users who are familiar with Big5 code input method (內碼輸入法) since the DOS era. They use traditional Bopomofo to input Hanzi but Big5 code for punctuations and symbols. Using big5 code has already become a muscle memory. The feature is for such users.

To enter Big5 code mode, just enter ctrl key and `. Type esc to leave the mode.


lukhnos (approved):

lukhnos:
LGTM :)

PR #372, opened 2023-10-01
System service to add to user phrases list /「加入小麥注音使用者詞彙」系統服務

shuang886:
### 使用方法：

1. 選擇某常用詞句
2. ⌘-點擊，選取「服務」→「加入小麥注音使用者詞彙」選項
3. 在編輯器裡檢視自動填入的注音

截圖 2023-10-01 下午3 57 22

### 注意：

開發時，系統可能需要強迫重讀新的服務才能使用：

```
    /System/Library/CoreServices/pbs -update
    /System/Library/CoreServices/pbs -flush
```

### 技術細節：

ParselessPhraseDB:
- search database for lines that match the "value" (phrase) field of the record
- 在內建詞庫裡搜尋符合一個詞的項目。這跟輸入法從注音找詞的用法相反。

McBopomofoLM, ParselessLM:
- pass the found records up
- 上傳找到的項目

LanguageModelManager:
- go through the records and find the highest-scoring match for a phrase, and return the "key" (pronounciation)
- 在符合的項目裡依分數排序，回覆最高分的注音。

AppDelegate, InputMethodController:
- move `openUserPhrases()` and others to AppDelegate for reuse, because InputMethodController is created by the system and there doesn't appear to be an easy way to get to it.
- 把 `openUserPhrases()` 等移到 AppDelegate 重用。InputMethodController 似乎沒有簡易取得物件的方式。

ServiceProvider:
- greedily match the longest possible substrings of the supplied string
- write the new entry to the user phrases list
- open the user phrases in an editor for the user to verify
- 在資料庫裡找詞，以最長的字串優先。
- 把新詞和注音寫入使用者詞彙。
- 啟動編輯器讓使用者檢查注音是否正確。

### 附註：

C++ 多年沒用，有點生疏了。有什麼錯誤請多包涵、指教。

### DIFF
--- a/Source/AppDelegate.swift
+++ b/Source/AppDelegate.swift
@@ -153,8 +153,8 @@ class AppDelegate: NSObject, NSApplicationDelegate, NonModalAlertWindowControlle
     private var updateNextStepURL: URL?
     private var fsStreamHelper: FSEventStreamHelper?
 
-    func updateUserPhases() {
-        NSLog("updateUserPhases called \(LanguageModelManager.dataFolderPath)")
+    func updateUserPhrases() {
+        NSLog("updateUserPhrases called \(LanguageModelManager.dataFolderPath)")
         LanguageModelManager.loadUserPhrases()
         LanguageModelManager.loadUserPhraseReplacement()
 
@@ -167,7 +167,7 @@ class AppDelegate: NSObject, NSApplicationDelegate, NonModalAlertWindowControlle
 
     func applicationDidFinishLaunching(_ notification: Notification) {
         LanguageModelManager.setupDataModelValueConverter()
-        updateUserPhases()
+        updateUserPhrases()
 
         if UserDefaults.standard.object(forKey: kCheckUpdateAutomatically) == nil {
             UserDefaults.standard.set(true, forKey: kCheckUpdateAutomatically)
@@ -175,8 +175,10 @@ class AppDelegate: NSObject, NSApplicationDelegate, NonModalAlertWindowControlle
         }
 
         NotificationCenter.default.addObserver(forName: .userPhraseLocationDidChange, object: nil, queue: OperationQueue.main) { notification in
-            self.updateUserPhases()
+            self.updateUserPhrases()
         }
+
+        NSApp.servicesProvider = ServiceProvider()
 
         checkForUpdate()
     }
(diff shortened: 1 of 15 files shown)


Source/AppDelegate.swift
```diff
@@ -153,8 +153,8 @@ class AppDelegate: NSObject, NSApplicationDelegate, NonModalAlertWindowControlle
     private var updateNextStepURL: URL?
     private var fsStreamHelper: FSEventStreamHelper?
 
-    func updateUserPhases() {
-        NSLog("updateUserPhases called \(LanguageModelManager.dataFolderPath)")
+    func updateUserPhrases() {
```
lukhnos, line 156: Thanks for fixing the typo here!

Source/Engine/McBopomofoLM.h
```diff
@@ -105,6 +105,12 @@ class McBopomofoLM : public Formosa::Gramambular2::LanguageModel {
     const std::vector<std::string> associatedPhrasesForKey(const std::string& key);
     bool hasAssociatedPhrasesForKey(const std::string& key);
 
+    /// Returns a list of pronounciations that match a given value.
+    /// @param value A string representing the text to look up pronounciation candidates for. For example,
+    ///     if you pass "說", it returns a list of records that include ㄕㄨㄛ, ㄕㄨㄟˋ, and ㄩㄝˋ.
+    /// @return An unsorted list of string records of the form "<key> <value> <score>" separated by spaces.
+    std::vector<std::string_view> getPronounciations(const std::string_view& value);
```
lukhnos: Typo: pro_nun_ciation (even though the verb is to pro_noun_ce…), and this will need a global search/replace.

Alternatively, consider just calling it "reading", so `getReadings`, which is the term McBopomofo uses in other places.

Source/LanguageModelManager.mm
```diff
@@ -305,4 +305,28 @@ + (void)setPhraseReplacementEnabled:(BOOL)phraseReplacementEnabled
     gLanguageModelMcBopomofo.setPhraseReplacementEnabled(phraseReplacementEnabled);
 }
 
++ (nullable NSString *)pronounciationFor:(NSString *)phrase {
```
lukhnos: Same here, pro_nun_ciation, but maybe also consider calling this `readingFor:` for the reason suggested in another comment.

Source/Engine/ParselessLM.cpp
```diff
@@ -144,3 +144,12 @@ bool McBopomofo::ParselessLM::hasUnigrams(const std::string& key)
 
     return db_->findFirstMatchingLine(key + " ") != nullptr;
 }
+
+std::vector<std::string_view> McBopomofo::ParselessLM::getPronounciations(const std::string_view& value)
```
lukhnos: Since `std::stirng_view` is basically a pointer, the returned vector doesn't have any lifetime guarantee. If the underlying string is gone, the returned vector will now contain dangling pointers. This will need to be a `std::vector<std::string>`.

But also see my comment on the call site of this method in LanguageModelManager.mm.

Source/LanguageModelManager.mm
```diff
@@ -305,4 +305,28 @@ + (void)setPhraseReplacementEnabled:(BOOL)phraseReplacementEnabled
...
+    if (!gLanguageModelMcBopomofo.isDataModelLoaded()) {
+        [self loadDataModel:InputModeBopomofo];
+    }
+
+    std::vector<std::string_view> records = gLanguageModelMcBopomofo.getPronounciations(phrase.UTF8String);
```
lukhnos: I already made a comment on the issue with using vector of string views, so I'm going to focus on the design issue here.

Namely, here the internal knowledge of the LM ("a record row has three columns") is leaked into the call site here. Wouldn't it be better if the parsing is done in ParselessLM? This will mean you'll want to define a data structure for reverse lookup though (how about a struct like `FoundReading` with two fields, `reading` and `score`?), but then this method can just go through the returned vector of the found readings without having to have the (leaky and potentially fragile) knowledge of splitting the record rows on its own.

Source/Engine/ParselessPhraseDB.h
```diff
@@ -50,6 +50,11 @@ class ParselessPhraseDB {
 
     const char* findFirstMatchingLine(const std::string_view& key);
 
+    // Find the rows that (prefix-)match the value, useful for returning all the
+    // ways a phrase or character can be pronounced. Note that this is a potentially-
```
lukhnos, line 54: No action required. I think this is fine. Linear-scanning 150k lines of data for a few times per system service invocation is acceptable for the use case of this PR.

lukhnos (changes requested):
I tried the PR, and it worked pretty well. I left a few comments, please take a look. Thank you for the proposed PR!

Source/Engine/McBopomofoLM.h
shuang886: Heh, I was telling myself the word looks wrong as I was typing them... thanks!

shuang886:
Thanks for the review. I've pushed fixes that should address all of your concerns.

lukhnos (approved):
LGTM. Thank you for the contribution!

PR #381, opened 2023-11-22
Add user default to allow candidate selection using numeric keypad

### LINKED ISSUE #345: [功能建議] 希望能透過數字鍵區的數字選字
**痛點**
1. 使用數字選字時，按下數字鍵區(九宮格區域)的數字會變成直接輸入該數字，而非選取該數字對應的候選字。

**功能說明**
1. 除了打字鍵區的數字，也能透過數字鍵區的數字選字。

**替代方案**
1. 繼續使用打字鍵區的數字選字。

**其他**
感謝各位辛苦的付出 :)

shuang886:
- default off, enable with "defaults write org.openvanilla.inputmethod.McBopomofo SelectCandidateWithNumericKeypad -bool YES"
- does not include Settings UI. I think it'd be a separate discussion as to whether UI is warranted.

fixes https://github.com/openvanilla/McBopomofo/issues/345

### DIFF
--- a/Source/KeyHandler.mm
+++ b/Source/KeyHandler.mm
@@ -266,7 +266,7 @@ InputMode InputModePlainBopomofo = @"org.openvanilla.inputmethod.McBopomofo.Plai
         return YES;
     }
 
-    if ([input isNumericPad]) {
+    if (!Preferences.selectCandidateWithNumericKeypad && [input isNumericPad]) {
         if (![input isLeft] && ![input isRight] && ![input isDown] && ![input isUp] && charCode != 32 && isprint(charCode)) {
             [self clear];
             InputStateEmpty *emptyState = [[InputStateEmpty alloc] init];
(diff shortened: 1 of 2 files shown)


Source/KeyHandler.mm
```diff
@@ -266,7 +266,7 @@ - (BOOL)handleInput:(KeyHandlerInput *)input state:(InputState *)inState stateCa
         return YES;
     }
 
-    if ([input isNumericPad]) {
+    if (!Preferences.selectCandidateWithNumericKeypad && [input isNumericPad]) {
```
lukhnos: How about switching the evaluation order here? That is, change the predicate to: `[input isNumericPad] && !Preferences.selectCandidateWithNumericKeypad`, so that `Preferences` is only checked when really needed, whereas right now it's always checked even if the key event is not relevant.

lukhnos (changes requested):
Thanks for the contribution! This LGTM, just one requested change.

Source/KeyHandler.mm
shuang886: https://github.com/openvanilla/McBopomofo/pull/381/commits/ec8d9ff1bea2cdc1f072fff7e209d35ea9548a16

lukhnos (approved):
Thanks!

PR #386, opened 2023-11-23
replace "Installation Successful" dialog with text in installer window

shuang886:
At least in Sonoma, both the installer and System Settings pop a dialog when the installation completes:

Screenshot 2023-11-23 at 7 16 34 AM Screenshot 2023-11-23 at 7 16 44 AM

which seems distracting. This proposal replaces the installer dialog with some more sedate:

Screenshot 2023-11-23 at 8 56 01 AM

### DIFF
--- a/Source/Installer/zh-Hant.lproj/MainMenu.xib
+++ b/Source/Installer/zh-Hant.lproj/MainMenu.xib
@@ -171,7 +171,6 @@ Gw
         <customObject id="494" customClass="AppDelegate">
             <connections>
                 <outlet property="cancelButton" destination="592" id="710"/>
-                <outlet property="installButton" destination="575" id="709"/>
                 <outlet property="progressIndicator" destination="4v9-yU-zek" id="W8M-2S-oXX"/>
                 <outlet property="progressSheet" destination="Pxv-O2-I1W" id="mp3-fA-PoS"/>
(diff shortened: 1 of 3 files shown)


Source/Installer/zh-Hant.lproj/MainMenu.xib
```diff
@@ -171,7 +171,6 @@ Gw
         <customObject id="494" customClass="AppDelegate">
             <connections>
                 <outlet property="cancelButton" destination="592" id="710"/>
```
lukhnos, line 175: the `@IBOutlet` connection to `actionButton` is missing in the zh-Hant version of the xib.

lukhnos (changes requested):
Good idea, thanks again for the contribution.

This looks good. Just one missing `@IBOutlet` connection in the zh-Hant xib.

Source/Installer/zh-Hant.lproj/MainMenu.xib
shuang886: Oh, rats. Thanks for the catch.

Fixed in https://github.com/shuang886/McBopomofo/commit/353702c4b58529807582ae8c085e2e1fe0b5291a

lukhnos (approved):
LGTM. Thanks again!

PR #387, opened 2023-11-24
Ensure default font size selected in Preferences

lukhnos:
This bug may have existed for a long time. In short the pop up button does not select the default candidate font size (16 pt) when McBopomofo is first installed. Since the user defaults are not yet populated, the pop up button cannot find a value to select, and so an empty item is added and selected instead. This commit fixes the issue.


zonble (commented):
LGTM. Thanks!

PR #383, opened 2023-11-23
update the look of the candidate sheet

shuang886:
- rounded rectangles
- honor light/dark modes

Screenshot 2023-11-22 at 7 29 09 PM
Screenshot 2023-11-22 at 7 28 55 PM
<br>
Screenshot 2023-11-22 at 7 35 24 PM
Screenshot 2023-11-22 at 7 35 14 PM

intended to fix https://github.com/openvanilla/McBopomofo/issues/310

PR #391, opened 2023-11-27
Honor the NSScroller.Style.overlay scrollbar style

lukhnos:
@shuang886 Thanks again for your PR #383. After that, I wonder if we should start honor the `.overlay` scrollbar style. May I ask for your opinion for this one?

Here's the problem with our currently forced use of `.legacy`:

current-ondemand

This is when "System Preferences > Appearance > Show scroll bars" is set to "Auto" (the default) or "When scrolling".

If the user sets to "Always" then things look fine:

current-always

In comparison, the retired candidate UI avoided laying the highlight on the scrollbar:

legacy

Now, one (not very rigorous) reason we wanted to always show the scrollbar was that we thought it gave a visual hint of "there's more", but I now think that the short appear-then-fade scrollbar when the table view has more than a pageful of content should suffice. macOS's built-in input methods also doesn't explicitly indicate "there's more" other than going along with the scrollbar convention.

After this commit, here's what the candidate sheet looks like (after the appear-and-fade of the scrollbar):

after-ondemand

When the apperance setting is set to Always, the look remains the same:

after-always

### DIFF
--- a/Packages/CandidateUI/Sources/CandidateUI/VerticalCandidateController.swift
+++ b/Packages/CandidateUI/Sources/CandidateUI/VerticalCandidateController.swift
@@ -327,13 +327,6 @@ extension VerticalCandidateController: NSTableViewDataSource, NSTableViewDelegat
             }
 
             keyLabelStripView.setNeedsDisplay(keyLabelStripView.frame)
-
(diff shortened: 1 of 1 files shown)


zonble (approved):

shuang886:
Makes sense to me, FWIW.

PR #392, opened 2023-11-28
Tunes the UI for the candidate 

zonble:
1. Change the color of the key caps from secondary label color to label color. When we are in the light mode but typing in an app with a dark background, such as Terminal app, it is hard to read the key caps in the secondary label color.
2. Also update the selected table row when scrolling the vertical candidate window.

### DIFF
--- a/Packages/CandidateUI/Sources/CandidateUI/VerticalCandidateController.swift
+++ b/Packages/CandidateUI/Sources/CandidateUI/VerticalCandidateController.swift
@@ -48,7 +48,7 @@ fileprivate class VerticalKeyLabelStripView: NSView {
 
         let textAttr: [NSAttributedString.Key: AnyObject] = [
             .font: keyLabelFont,
-            .foregroundColor: NSColor.secondaryLabelColor,
+            .foregroundColor: NSColor.labelColor,
             .paragraphStyle: paraStyle]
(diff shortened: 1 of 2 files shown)


Packages/CandidateUI/Sources/CandidateUI/VerticalCandidateController.swift
```diff
@@ -134,6 +134,7 @@ public class VerticalCandidateController: CandidateController {
         scrollView.verticalScrollElasticity = .none
         scrollView.drawsBackground = false
         scrollView.contentView.drawsBackground = false
+        scrollView.contentView.postsBoundsChangedNotifications = true
 
         tableView = NSTableView(frame: contentRect)
```
lukhnos, line 139: I was wondering if using the `VerticalCandidateTableView` implemented in this file (which has a `adjustScroll` override used in the old, Objective-C version) instead of `NSTableView` would solve the problem, but then I've tested both, and I like the new solution a lot. The legacy solution would effectively "add a lot of resistance" to the scroll (since the user needs to scroll past the row height threshould), whereas this solution allows the user to scroll as few points as they like and then causes the key strip label highlight to catch up. So this LGTM.

lukhnos (approved):

PR #399, opened 2023-12-11
Introduces input macros

zonble:
The PR introduces input macros. Input macros are candidates that would be
replace dynamically by the input engine. We use input macros to input date and
time since a user might be hardly to remember the current date (especially in
the traditional Chinese calendar such as 癸卯年).

Currently, it supports the following input macros:

| 輸入     | 範例                    |
| -------- | ----------------------- |
| 今天日期 | 2023 年 12 月 11 日     |
| 明天日期 | 民國 112 年 12 月 12 日 |
| 現在時間 | 下午 2:28:21            |
| 所在時區 | 台北標準時間            |
| 今年干支 | 癸卯年                  |
| 今年生肖 | 水兔年                  |

The localized dates are powered by ICU, so ICE becomes a dependency.


Source/InputMacro.swift
```diff
@@ -0,0 +1,326 @@
...
+    let year = calendar.component(.year, from: date)
+    return year
+}
+
+fileprivate func getYearBase(year: Int) -> (Int, Int) {
+    let base: Int = year < 4 ? 60 - ((year * -1) + 2) % 60 : (year - 3) % 60
+    return (base % 10, base % 12)
+}
+
+fileprivate func ganshi(year: Int) -> String {
```
lukhnos: Nit: ganzhi ?

Source/InputMacro.swift
```diff
@@ -0,0 +1,326 @@
...
+    }
+}
+
+// MARK: - InputMacroController
+
+class InputMacroController: NSObject {
+    @objc
+    static let shared = InputMacroController()
+
+    private var macros: [InputMacro] =
```
lukhnos: Would it be better to create a dictionary from this array during init, and let `handle` look up the dictionary?

In `McBopomofoLM`, the macro converter is invoked for every unigram fetched. So for example if we type "今天日期", at least five unigrams ("今天日期" plus the other four `MACRO@` macro values) are retrieved, and all five of them have to pass through this sequential lookup.

lukhnos (changes requested):

Source/InputMacro.swift
zonble: OK. Let me do this.

Source/InputMacro.swift
zonble: Yap.. it is a typo.

lukhnos (approved):

PR #406, opened 2023-12-25
Introduces a key to look up a candidate in various dictionaries

zonble:
The PR helps users to know about the detailed information about a candidate. When a user opens the candidate window to list all available candidates, he or she can highlight a candidate and press on the "?" key. A menu will be opened to list available dictionaries.

Screenshot 2023-12-26 at 12 14 17 AM

Screenshot 2023-12-26 at 12 14 23 AM

Then, the user can pick a dictionary service from the menu, and McBopomofo will open external web browser, dictionary app and so on accordingly. One of the options is to let the user to see the character code of the highlighted candidate.

Screenshot 2023-12-26 at 12 14 36 AM


Packages/CandidateUI/Sources/CandidateUI/VerticalCandidateController.swift
```diff
@@ -281,6 +281,10 @@ public class VerticalCandidateController: CandidateController {
                 newIndex = 0
             }
 
+            if newIndex == UInt.max {
```
lukhnos, line 284: I think the same commit to OpenVanilla should be carried over here too. That is, a `UInt.max` check should be placed in line 297, and these three lines (284-286 currently) are dead code anyway, because `newIndex` is always set to 0 if `newValue` is `UInt.max` (from lines 276-282).

Source/zh-Hant.lproj/Localizable.strings
```diff
@@ -96,3 +96,20 @@
...
+
+"Look up \"%1$@\" in %2$@" = "在%2$@查詢「%1$@」";
+
+"Dictionary app" = "字典App";
+
+"MOE Dict" = "萌典";
+
+"MOE Dict (Holo)" = "台語萌典";
+
+"MOE Dict (Holo)" = "客語萌典";
```
lukhnos: nit: `"MOE Dict (Hakka)"` instead?

Source/en.lproj/Localizable.strings
```diff
@@ -96,3 +96,18 @@
...
 
 "McBopomofo is up to date." = "McBopomofo is up to date.";
+
+"Look up \"%1$@\" in %2$@" = "Look up \"%1$@\" in %2$@";
+
+"Dictionary app" = "Dictionary app";
+
+"MOE Dict" = "MOE Dict";
+
+"MOE Dict (Holo)" = "MOE Dict (Holo)";
```
lukhnos: nit: Missing entry for `"MOE Dict (Hakka)"`?

Source/dictionary_service.json
```diff
@@ -0,0 +1,60 @@
+{
+  "services": [
+    {
+      "name": "Dictionary app",
+      "url_template": "dict://(encoded)"
+    },
+    {
+      "name": "MOE Dict",
```
lukhnos: I was wondering if this should be called `"moedict.tw"` since "MOE Dict" sounds like the officical MOE dictionary. Or should this simply be called "萌典" without localization just like "教育部重編國語詞典修訂本" below?

Calling the three entries 萌典、萌典（台語）、萌典（客語） also removes the need to localize them in English.

lukhnos (approved):
LGTM! Just a few comments on the service entry JSON and whether those need to be localized at all. Feel free to merge.

lukhnos (approved):
Thanks!

PR #412, opened 2024-01-01
Introduces associated phrases in the smart bopomofo mode

zonble:
The associated phrases is a feature lives in the traditional phonetic input methods, as the "Plain Bopomofo" mode of McBopomofo, or ㄅ半. The PR tries to bridge the feature to the smart bopomofo mode.

When a user presses shift + enter after inputting a character, the input method will open a menu containing associated phrases, then it inserts the readings and fixes the node after the user selects one. 

Or, the user can press shift + enter on a candidate when the candidate window appears, and the input method will give the second level of the candidate window of the associated phrases. When a phrase is selected, the input method selects the candidate and inserts the readings accordingly.

A user can use esc to cancel the associated phrases candidate window anytime.


lukhnos (changes requested):
This is looking good. Could you cherry-pick https://github.com/openvanilla/McBopomofo/commit/008bdcfd59685907c75fccc00da779c355b9ab9d ? I made some changes to `reading_grid` so that you don't have to comment out the `protected:` to expose the methods you need. I also removed some debug logging in the key handler.

zonble:
008bdcfd59685907c75fccc00da779c355b9ab9d  was picked :)

lukhnos (approved):

PR #417, opened 2024-01-07
Creates a menu to do additional inputting

zonble:
The PR lets the users to be able to show a menu by pressing Ctrl + \. The menu contains following features

- Lowered cases Chinese numbers: 12349876 will become 一千二百三十四萬九千八百七十六
- Uppercases Chinese numbers, such as 壹仟貳佰參拾肆萬玖仟捌佰柒拾陸
- Suzhou numbers (蘇州碼 or 花碼). For example, 33.45 will become

```
〣三〤〥
十[單位]
```
- Other existing features like Big5 code and date macros.


zonble:
Video demo


https://github.com/openvanilla/McBopomofo/assets/53011/997ce7eb-01c4-4a8a-96e3-699cdbc720ab

Source/zh-Hant.lproj/Localizable.strings
```diff
@@ -105,4 +105,15 @@
...
-"Look up %@" = "Look up %@";
+"Look up %@" = "查找 %@";
+
+"Date and Time" = "日期與時間";
+
+"Big5 Code" = "Big5 內碼輸入";
+
+"Lower Case Chinese Numbers" = "小寫中文數字";
+
+"Upper Case Chinese Numbers" = "大小中文數字";
```
ChiahongHong: Typo: `大小中文數字` should be `大寫中文數字`

Source/en.lproj/Localizable.strings
```diff
@@ -108,3 +108,13 @@
 "%@ has been copied." = "%@ has been copied.";
 
 "Look up %@" = "Look up %@";
+
+"Date and Time" = "Date and Time";
+
+"Big5 Code" = "Big5 Code";
+
+"Lower Case Chinese Numbers" = "Lower Case Chinese Numbers";
```
lukhnos: nit: s/Lower Case/Lowercase/g, also it should be "Uppercase"

Source/Base.lproj/Localizable.strings
```diff
@@ -106,3 +106,13 @@
 "%@ has been copied." = "%@ has been copied.";
 
 "Look up %@" = "Look up %@";
+
+"Date and Time" = "Date and Time";
+
+"Big5 Code" = "Big5 Code";
+
+"Lower Case Chinese Numbers" = "Lower Case Chinese Numbers";
```
lukhnos: "Lowercase" and "Uppercase", same for other places.

Packages/ChineseNumber/Sources/ChineseNumber/ChineseNumber.m
```diff
@@ -0,0 +1,32 @@
+#import "ChineseNumber.h"
+
+const char *chinese_number(const char *int_part, const char *dec_part, bool use_formal);
+extern const char *suzhou_number(const char *int_part, const char *dec_part, bool vertical_digit_at_start, const char *unit);
```
lukhnos: I don't think you need the `extern` here.

Packages/ChineseNumber/Sources/ChineseNumber/ChineseNumber.m
```diff
@@ -0,0 +1,32 @@
...
+
+@implementation ChineseNumber
+
++ (NSString *)lowerNumberWithIntPart:(NSString *)intPart decPart:(NSString *)decPart
+{
+    const char *result = chinese_number (intPart.UTF8String, decPart.UTF8String, false);
```
lukhnos: Hmm, C string processing is hard. Ever considered implementing this in pure Objective-C (so using `NSString` only) or C++ `std::string`? There are some issues with buffer size and buffer overflow (even if rare, practically, but still real issues nonetheless) here, and the error handling and edge cases may require much more code than if written in string-safe Objective-C or C++. See my comments below…

Source/KeyHandler.mm
```diff
@@ -607,6 +611,22 @@ - (BOOL)handleInput:(KeyHandlerInput *)input state:(InputState *)inState stateCa
         }
     }
 
+    NSLog(@"char code %d", charCode);
```
lukhnos: Nit: remove logging here.

Packages/ChineseNumber/Sources/ChineseNumber/chinese_number.m
```diff
@@ -0,0 +1,92 @@
...
+
+static char *lower_digits[10] = {"〇", "一", "二", "三", "四",
+        "五", "六", "七", "八", "九"};
+static char *upper_digits[10] = {"零", "壹", "貳", "參", "肆",
+        "伍", "陸", "柒", "捌", "玖"};
+static char *lower_places[4] = {"千", "百", "十", ""};
+static char *upper_places[4] = {"仟", "佰", "拾", ""};
+static char *higher_place_names[10] = {"", "萬", "億", "兆", "京", "垓", "秭", "穰"};
+
+const char *convert_4_four_digits_to_chinese(const char *input,
```
lukhnos: Perhaps doc-comment on every C function introduced in this PR that the caller is responsible for freeing the returned string?

Packages/ChineseNumber/Sources/ChineseNumber/chinese_number.m
```diff
@@ -0,0 +1,92 @@
...
+                strcat(buffer, is_uppercase ? upper_digits[0] : lower_digits[0]);
+            }
+            handle_zero_on_next_digit = true;
+            zero_happened = false;
+            strcat(buffer, is_uppercase ? upper_digits[c - '0'] : lower_digits[c - '0']);
+            strcat(buffer, is_uppercase ? upper_places[i] : lower_places[i]);
+        }
+    }
+
+    return buffer;
```
lukhnos: For this particular function, the longest possible output is `九千九百九十九`, which is 22 bytes including `\0`, so I'd begrudingly say this is ok…

Packages/ChineseNumber/Sources/ChineseNumber/chinese_number.m
```diff
@@ -0,0 +1,92 @@
...
+        strncpy(part, filled_int_part + i, 4);
+        if (part[0] == '0' && part[1] == '0' && part[2] == '0' && part[3] == '0') {
+            i += 4;
+            zero_happened = true;
+            continue;
+        }
+
+        char *converted = (char *) convert_4_four_digits_to_chinese (part, i == 0, zero_happened, use_formal);
+        zero_happened = false;
+        strcat(output, (const char *) converted);
```
lukhnos: For the five `strcat` calls in this function, there's the risk of buffer overflow in `output` (even though it'd require a very large user input). What happens when `strlen(output) + strlen(converted) + 1 > 2048`?

Packages/ChineseNumber/Sources/ChineseNumber/chinese_number.m
```diff
@@ -0,0 +1,92 @@
...
+    const char *trimmed_int_part = trim_zero (int_part, true);
+    const char *trimmed_dec_part = trim_zero (dec_part, false);
+
+    size_t four_digits_sets_count = (strlen (trimmed_int_part) / 4);
+    if (strlen (trimmed_int_part) % 4 != 0) {
+        four_digits_sets_count++;
+    }
+    size_t filled_length = four_digits_sets_count * 4;
+    const char *filled_int_part = left_padding_with_zero (trimmed_int_part, filled_length);
+    char *output = calloc (2048, 1);
```
lukhnos: Nit: use macro/const for magic number?

Packages/ChineseNumber/Sources/ChineseNumber/suzhou_number.m
```diff
@@ -0,0 +1,111 @@
...
+            "億", "十億", "百億", "千億",
+            "兆", "十兆", "百兆", "千兆",
+            "京", "十京", "百京", "千京",
+            "垓", "十垓", "百垓", "千垓",
+            "秭", "十秭", "百秭", "千秭",
+            "穰", "十穰", "百穰", "千穰",
+    };
+    struct suzhou_number_struct data = build_struct (int_part, dec_part);
+    char *output = calloc (1024, 1);
+    memset((void *) output, 0, BUFFER_SIZE);
```
lukhnos: Inconsistent buffer size here. `output` is allocated 1024 bytes (also a magic number), but `BUFFER_SIZE` is defined to be 80.

Packages/ChineseNumber/Sources/ChineseNumber/suzhou_number.m
```diff
@@ -0,0 +1,111 @@
...
+    char joined[BUFFER_SIZE] = "";
+    strcat(joined, data.int_part);
```
lukhnos: What happens when `data.int_part` or `data.dec_part` is longer that `joined`?

Packages/ChineseNumber/Sources/ChineseNumber/suzhou_number.m
```diff
@@ -0,0 +1,111 @@
...
+    strcat(joined, data.dec_part);
+
+    bool use_vertical = vertical_digit_at_start;
+    size_t code_point_count = 0;
+    for (size_t i = 0; i < strlen (joined); i++) {
+        char c = (char) (joined[i] - '0');
+        if (c >= 1 && c <= 3) {
+            if (use_vertical) {
+                use_vertical = false;
+                strcat(output, vertical_digits[c]);
```
lukhnos: Same general problem with `chinese_number.m`, where `strcat` is called without checking whether `output` has enough space.

Packages/ChineseNumber/Sources/ChineseNumber/trim_zero.m
```diff
@@ -0,0 +1,55 @@
+#include "trim_zero.h"
+#include <stdlib.h>
+#include <string.h>
+
+#define BUFFER_SIZE 80
+
+const char *trim_zero(const char *input, bool from_start) {
+    char *rtn = calloc (BUFFER_SIZE, 1);
```
lukhnos: What happens if `input` is longer than `BUFFER_SIZE`?

Packages/ChineseNumber/Sources/ChineseNumber/trim_zero.m
```diff
@@ -0,0 +1,55 @@
...
+            char c = input[i];
+            if (non_zero_found) {
+                rtn[n++] = c;
+            } else if (c != '0') {
+                non_zero_found = true;
+                rtn[n++] = c;
+            }
+        }
+    } else {
+        char reversed[BUFFER_SIZE] = "";
```
lukhnos: `rtn` is dynamically allocated but `reverse` is stack-allocated.

Packages/ChineseNumber/Sources/ChineseNumber/trim_zero.m
```diff
@@ -0,0 +1,55 @@
...
+        }
+    }
+    return rtn;
+}
+
+const char *left_padding_with_zero(const char *input, size_t target) {
+    char *buffer = (char *) calloc (BUFFER_SIZE, 1);
+    memset(buffer, 0,  BUFFER_SIZE);
+    size_t len = strlen (input);
+    if (len >= target) {
```
lukhnos: Neither the size of `input` nor `target` is checked against buffer length.

Packages/ChineseNumber/Sources/ChineseNumber/suzhou_number.m
```diff
@@ -0,0 +1,111 @@
...
+    size_t int_trimmed_zero_count = 0;
+    if (!strlen (initial_dec_part)) {
+        int_trimmed_zero_count = strlen (initial_int_part) - strlen (trimmed_int_part);
+    } else {
+        free ((void *) trimmed_int_part);
+        trimmed_int_part = initial_int_part;
+    }
+
+    if (!strlen (trimmed_int_part)) {
+        memcpy((void *) trimmed_int_part, (void *) "0", 2);
```
lukhnos: This is only "safe" because from the implementation detail we know that `trim_zero` always allocates a block of size `BUFFER_SIZE`, but normally this `strlen`-`memcpy` operation is best avoided…

Packages/ChineseNumber/Sources/ChineseNumber/suzhou_number.m
```diff
@@ -0,0 +1,111 @@
+#include <string.h>
+#include <stdlib.h>
+#include "trim_zero.h"
+
+#define BUFFER_SIZE 80
+
+struct suzhou_number_struct
```
lukhnos: The problem with having C strings in a struct is that it's hard to remember to free all the buffers allocated here. This is a problem one wouldn't have if this were written with `std::string`s…

Source/KeyHandler.mm
```diff
@@ -607,6 +611,22 @@ - (BOOL)handleInput:(KeyHandlerInput *)input state:(InputState *)inState stateCa
...
+
+    if ([input isControlHold] && (charCode == 28)) {
```
lukhnos: Perhaps comment which key combo this is. The PR says the activation key is `Ctrl + .`, but 28 is actually ASCII `FS` and the key combo is (on my keyboard) `Ctrl + \`…

lukhnos (commented):
I enjoyed the feature addition a lot. Overall this looks good. My biggest concern though is the C string operations introduced in the PR, and I've commented where relevant. PTAL.

Packages/ChineseNumbers/Sources/ChineseNumbers/SuzhouNumbers.swift
```diff
@@ -0,0 +1,71 @@
...
+        "0": "〇", "1": "〡", "2": "〢", "3": "〣", "4": "〤",
+        "5": "〥", "6": "〦", "7": "〧", "8": "〨", "9": "〩"
+    ]
+    static private let horizontalDigits: [Character: String] = [
+        "1": "一", "2": "二", "3": "三"
+    ]
+    static private let placeNames = [
+        "", "十", "百", "千",
+        "万", "十万", "百万", "千万",
+        "億", "十億", "百億", "千億",
```
tianjianjiang, line 17: Are 万's deliberately in Simplified Chinese? If the intention is to support usages in China, then as far as I can tell, people from there use 万億 instead of 兆 nowadays.

Source/Base.lproj/Localizable.strings
zonble: Fixed.

Source/en.lproj/Localizable.strings
zonble: Fixed

Source/KeyHandler.mm
zonble: Fixed.

Packages/ChineseNumber/Sources/ChineseNumber/trim_zero.m
zonble: The library is now in Swift :)

Packages/ChineseNumber/Sources/ChineseNumber/trim_zero.m
zonble: The library is now in Swift :)

Packages/ChineseNumber/Sources/ChineseNumber/ChineseNumber.m
zonble: The library is now in Swift :)

Source/zh-Hant.lproj/Localizable.strings
zonble: Fixed.

Packages/ChineseNumber/Sources/ChineseNumber/suzhou_number.m
zonble: The library is now in Swift :)

Packages/ChineseNumbers/Sources/ChineseNumbers/SuzhouNumbers.swift
zonble: AFAIK, people use 万 when doing 苏州码. It is part of the ancient number system, and it is not about Simplified or Traditional Chinese. Actually I do not think people may use 苏州码 to present numbers larger than 億 but I just added it since I allowed users to input such a big number.

Packages/ChineseNumbers/Sources/ChineseNumbers/SuzhouNumbers.swift
```diff
@@ -0,0 +1,73 @@
+import Foundation
+
+public class SuzhouNumbers: NSObject {
+
+    static private let verticalDigits: [Character: String] = [
+        "0": "〇", "1": "〡", "2": "〢", "3": "〣", "4": "〤",
+        "5": "〥", "6": "〦", "7": "〧", "8": "〨", "9": "〩"
+    ]
+    static private let horizontalDigits: [Character: String] = [
+        "1": "一", "2": "二", "3": "三"
```
lukhnos, line 12: This is missing the mapping `"0": "〇"`, and so if you try to convert the number `1010` to Suzhou Number, it will crash the lookup at the current line 45: `horizontalDigits[c]!`.

Consider also add a unit test case for this (though we aren't running the ChineseNumbersTests as part of the CI workflow, are we)?

lukhnos (changes requested):

Packages/ChineseNumbers/Sources/ChineseNumbers/SuzhouNumbers.swift
tianjianjiang: I see. It makes sense to me, too, since 万 isn't necessarily China's simplified Chinese but rather a variant that is easier to write and not having strange and irrelevant origin like 萬.

BTW, not a big deal, but I can still find at least one picture with 萬: https://www.wenweipo.com/a/202110/06/AP615cb727e4b08d3407de5b43.html

Packages/ChineseNumbers/Sources/ChineseNumbers/SuzhouNumbers.swift
zonble: Fixed.

tianjianjiang (approved):

PR #421, opened 2024-01-20
Skip the test with Chinese TTS

zonble:
It looks like GitHub's macOS VMs are without Meija, the Chinese TTS voice, installed, and it caused our tests to fail. I just skip the test.


lukhnos (approved):
LGTM

PR #428, opened 2024-01-30
Sync engine code with fcitx5-mcbopomofo

lukhnos:
Also updates LanguageModelManager to match the updated function signatures.


zonble (approved):
LGTM.

PR #432, opened 2024-02-09
Makes the behavior of moving the cursor much like MS Bopomofo

zonble:
The details are discuessed in #431.

### DIFF
--- a/Source/KeyHandler.mm
+++ b/Source/KeyHandler.mm
@@ -535,7 +535,13 @@ InputMode InputModePlainBopomofo = @"org.openvanilla.inputmethod.McBopomofo.Plai
         }
 
         size_t originalCursorIndex = _grid->cursor();
-        if (originalCursorIndex == _grid->length() && Preferences.selectPhraseAfterCursorAsCandidate) {
+
+        // Note: When the cursor is at the end of the composing buffer and the
+        // preference that make McBopomofo be like MS Bopomofo are on, the
+        // cursor should be moved to the begin of the last character.
+        if (originalCursorIndex == _grid->length() &&
+            Preferences.selectPhraseAfterCursorAsCandidate &&
+            Preferences.moveCursorAfterSelectingCandidate) {
             _grid->setCursor(originalCursorIndex - 1);
         }
         InputStateChoosingCandidate *choosingCandidates = [self _buildCandidateState:(InputStateNotEmpty *)state useVerticalMode:input.useVerticalMode];
--- a/Source/KeyHandler.mm
+++ b/Source/KeyHandler.mm
@@ -1245,6 +1245,9 @@ InputMode InputModePlainBopomofo = @"org.openvanilla.inputmethod.McBopomofo.Plai
             InputStateEmptyIgnoringPreviousState *empty = [[InputStateEmptyIgnoringPreviousState alloc] init];
             stateCallback(empty);
         } else {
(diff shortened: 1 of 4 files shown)


Source/KeyHandler.mm
```diff
@@ -531,7 +533,19 @@ - (BOOL)handleInput:(KeyHandlerInput *)input state:(InputState *)inState stateCa
...
+
+        size_t originalCursorIndex = _grid->cursor();
+
+        // Note: When the cursor is at the end of the composing buffer and the
+        // preference that make McBopomofo be like MS Bopomofo are on, the
+        // cursor should be moved to the begin of the last character.
+        if (originalCursorIndex == _grid->length() &&
+            Preferences.selectPhraseAfterCursorAsCandidate &&
+            Preferences.moveCursorAfterSelectingCandidate) {
+            _grid->setCursor(originalCursorIndex - 1);
```
lukhnos, line 545: What would happen if `originalCursorIndex` is 0?

Source/KeyHandler.mm
```diff
@@ -635,7 +649,12 @@ - (BOOL)handleInput:(KeyHandlerInput *)input state:(InputState *)inState stateCa
                 [self _walk];
                 InputStateInputting *inputting = (InputStateInputting *)[self buildInputtingState];
                 stateCallback(inputting);
+                size_t originalCursorIndex = _grid->cursor();
+                if (Preferences.selectPhraseAfterCursorAsCandidate) {
```
lukhnos, line 653: Shouldn't the predicate be the same as that in lines 542-544? Otherwise this would always move the cursor back by 1 under any circumstances (that is, even if the cursor is not at the end of the buffer). Also, what happens if `originalCursorIndex` is already 0?

lukhnos (changes requested):

Source/KeyHandler.mm
zonble: Since originalCursorIndex is now as the length of the grid, and it cannot enter the candidate state if the length of the grid is zero, I do not think originalCursorIndex could be zero.

Source/KeyHandler.mm
zonble: This actually fixes a bug. We let the user to show the candidate window for the symbols by pressing the "\`" key, and what we do here is to insert a reading and then enter the candidate state. The bug was, if a user entered "你好", and then input "`" between "你" and "好", it will show the candidates of "好" but not the symbols. Thus, I move the cursor backward, and it will be "你|　好" instead of ""你　|好".

Since we insert a reading in the previous step, the cursor is moved forward by 1 step by the action of inserting the reading, and the cursor will not be 0.

Source/KeyHandler.mm
lukhnos: Makes sense. Thanks.

Source/KeyHandler.mm
lukhnos: I see. Thanks for the clarification!

lukhnos (approved):

Source/KeyHandler.mm
```diff
@@ -635,7 +649,12 @@ - (BOOL)handleInput:(KeyHandlerInput *)input state:(InputState *)inState stateCa
...
+                 _grid->setCursor(originalCursorIndex - 1);
```
lukhnos: nit: wrong indent on this line.

PR #437, opened 2024-02-24
Add missing phrases for all 47 prefectures and 9 regions of Japan

xatier:
* fixed the incorrect pronunciation of 潟 ㄒㄧˋ

Ref:

https://www.moedict.tw/%E6%BD%9F
https://www.wikiwand.com/zh-tw/%E9%83%BD%E9%81%93%E5%BA%9C%E7%B8%A3

----

I noticed some Japanese prefectures names are missing/inconsistent in the dictionary, I hence wrote a script to add the missing ones (for both `pref.` and `pref. 縣`). I tried my best to follow the orders and frequencies from the existing files, but I found both `BPMFMappings.txt ` and `phrase.occ` are somewhat unordered.


xatier:
On the ordering issue, I have this little script locally to keep my user dictionary nice and neat, I would suggest having sometime similar in this repo.

https://gist.github.com/xatier/9f34fb64884e3cbfd66f24716beb6e7a

tianjianjiang (approved):
LGTM, and some minor comments/questions below.


> * fixed the incorrect pronunciation of 潟 ㄒㄧˋ

Thank you for that!

---
> [...] add the missing ones (for both `pref.` and `pref. 縣`).

Is it because some name syllables followed by /ㄒㄧㄢˋ/　will form other Chinese characters? Any example? Just curious. The same question can go to 地方. As far as I can tell, Japanese 地方 will usually be correct, unless we want to have a Japanese kanji version of 中国地方 that separates from the cases of 中國 and 中囯.

---
> [...] `BPMFMappings.txt` and `phrase.occ` are somewhat unordered.
> [...] suggest having sometime similar in this repo.
> https://gist.github.com/xatier/9f34fb64884e3cbfd66f24716beb6e7a

I guess you may have noticed the importance of row order for `BPMFBase.txt` from [詞庫開發說明#修正讀音](https://github.com/openvanilla/McBopomofo/wiki/%E8%A9%9E%E5%BA%AB%E9%96%8B%E7%99%BC%E8%AA%AA%E6%98%8E#%E4%BF%AE%E6%AD%A3%E8%AE%80%E9%9F%B3}
However, row order isn't critical in `BPMFMappings.txt` and `phrase.occ`. Nonetheless, it is still nice to have for readability.

xatier:
@zonble thanks for merging the PR promptly.

@tianjianjiang 

1) 潟 is correct in `BPMFBase.txt` [1]; however, it was incorrect in `BPMFMappings.txt`.

```text
潟 ㄒㄧˋ xi4 vu4 big5
```

```
-新潟 ㄒㄧㄣ ㄒㄧㄝˋ
+新潟 ㄒㄧㄣ ㄒㄧˋ
```

2) I believe we only need to output Traditional Chinese characters, as `McBopomofo` is a Chinese input method, we don't need to output Japanese Kanji, unless we have some other use cases to justify that. My main concern was that I wasn't able to type `熊本` as a default phrase with `fcitx5-mcbopomofo` :) 

3) I noticed `BPMFMappings.txt` and `phrase.occ` are a bit unordered by various chunks, as I mentioned earlier. Hence, I proposed having some scripts to make the ordering deterministic, possibly also integrate that into the CI.

[1] https://github.com/openvanilla/McBopomofo/blob/master/Source/Data/BPMFBase.txt#L11670

tianjianjiang:
Hi @xatier,
> 1. 潟 [...] was incorrect in `BPMFMappings.txt`. [...]

Got it! Thanks for the further info.

> 2. I believe we only need to output Traditional Chinese characters, as `McBopomofo` is a Chinese input method, we don't need to output Japanese Kanji, unless we have some other use cases to justify that. My main concern was that I wasn't able to type `熊本` as a default phrase with `fcitx5-mcbopomofo` :)

I don't have any preference for Kanji, so we are on the same page, I believe.
And duly noted for the case of "`熊本`." My main reasoning is that, while `??縣` is necessary for not forming `??現`, it is probably ok to not have `??地方`. But for this case I don't have strong preference, either.

> 3. I noticed `BPMFMappings.txt` and `phrase.occ` are a bit unordered by various chunks, as I mentioned earlier. Hence, I proposed having some scripts to make the ordering deterministic, possibly also integrate that into the CI.

For readability, especially for diff, it sounds good to me. IMO, it's just a matter of priority. :)

PR #438, opened 2024-02-24
Taiwanese Braille support

zonble:
The PR adds following features

- A service menu to convert selected Chinese text to BPMF
- A service menu to convert selected Chinese text to Taiwanese Braille
- A service menu to convert Taiwanese Braille to Chinese Text
- A preference setting to let user to use Ctrl + Enter to convert current composing buffer to Taiwanese Braille

Also

- VXHanConvert is removed
- Adds additional emojis
- Users can now move the cursor with shift and arrow keys even the JK key option is not on when choosing candidates


lukhnos (approved):
LGTM. Thanks!

I don't think we have not documented how to customize the Ctrl-Enter behavior via setting the `ControlEnterOutput` preferences key. Should we update the wiki after this?

zonble:
@lukhnos I will update the preference UI later.

PR #424, opened 2024-01-29
Makes it able to move cursor when the candidate window is presented

zonble:
This PR fixes #422 and #423.

McBopomofo allows to set size per page to at least 4 candidates, however, there was a bug that causes the minimal size to be 5. The PR fixed the bug.

The PR also adds an option to let users to use J and K keys to move the cursor when the candidate window is visible.

PR #440, opened 2024-02-26
Fixes the cursor index in the choosing candidate state

zonble:
The way we set the cursor index in the candidate state was changed, and it caused wrong results. The PR fixed the issue.

Also 

- Update the preference UI for Ctrl+Enter key
- Make it able to let the macOS version to output HTML ruby

### DIFF
--- a/Source/KeyHandler.mm
+++ b/Source/KeyHandler.mm
@@ -598,13 +598,30 @@ InputMode InputModePlainBopomofo = @"org.openvanilla.inputmethod.McBopomofo.Plai
     // MARK: Enter
     if (charCode == 13) {
         if (_inputMode == InputModeBopomofo && input.isControlHold) {
-            if (Preferences.controlEnterOutput == ControlEnterOutputBpmfReading) {
-                return [self _handleCtrlEnterBpmfReadingWithState:state stateCallback:stateCallback errorCallback:errorCallback];
+            NSString *string = @"";
+            if (Preferences.controlEnterOutput == ControlEnterOutputOff) {
+                errorCallback();
+                return YES;
             }
-            if (Preferences.controlEnterOutput == ControlEnterOutputBraille) {
-                return [self _handleCtrlEnterBrailleWithState:state stateCallback:stateCallback errorCallback:errorCallback];
+            switch (Preferences.controlEnterOutput) {
+                case ControlEnterOutputBpmfReading:
+                    string = [self _currentBpmfReading];
+                    break;
+                case ControlEnterOutputHtmlRuby:
+                    string = [self _currentHtmlRuby];
+                    break;
+                case ControlEnterOutputBraille:
+                    string = [self _currentBraille];
+                    break;
+                default:
+                    break;
             }
-            errorCallback();
+            [self clear];
+
+            InputStateCommitting *committing = [[InputStateCommitting alloc] initWithPoppedText:string];
+            stateCallback(committing);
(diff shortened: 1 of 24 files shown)


Source/KeyHandler.mm
```diff
@@ -1860,8 +1860,9 @@ - (InputStateChoosingCandidate *)_buildCandidateState:(InputStateNotEmpty *)curr
         InputStateCandidate *candidate = [[InputStateCandidate alloc] initWithReading:r value:v displayText:dt];
         [candidatesArray addObject:candidate];
     }
+    InputStateInputting *inputting = (InputStateInputting *)[self buildInputtingState];
 
-    InputStateChoosingCandidate *state = [[InputStateChoosingCandidate alloc] initWithComposingBuffer:currentState.composingBuffer cursorIndex:_grid->cursor() candidates:candidatesArray useVerticalMode:useVerticalMode];
+    InputStateChoosingCandidate *state = [[InputStateChoosingCandidate alloc] initWithComposingBuffer:inputting.composingBuffer cursorIndex:inputting.cursorIndex candidates:candidatesArray useVerticalMode:useVerticalMode];
```
lukhnos, line 1883: Let's take a step back. The change to use `_grid->cursor()` (and hence the issue) was introduced in https://github.com/openvanilla/McBopomofo/pull/424/commits/e3b559594f8556cd13587a4805533f1c363adb6e because we mutated the grid's cursor to accomodate the new logic with the MS IME cursor mode. Could we do one of the following, either:

(1) Make it clear that the method `_buildCandidateState:useVerticalMode:` will always disregard the current state the IME is in, and will always build an inputting state on its own? If you take this path, you might as well removed the `currenState` parameter from this method. Or,

(2) Rebuild an inputting state after the cursor is mutated ([here](https://github.com/openvanilla/McBopomofo/blob/ae4146f3246ef230727834d163a8c518d686c4d1/Source/KeyHandler.mm#L545) and [here](https://github.com/openvanilla/McBopomofo/blob/ae4146f3246ef230727834d163a8c518d686c4d1/Source/KeyHandler.mm#L661), and revert this argument to `currentState.cursorIndex`?

I personally perfer (2)–I believe we wrote `_buildCandidateState:useVerticalMode:` with the intent that it returns a ChoosingCandidate state that _inherits_ the properties from `currentState`. Rebuilding an inputting state after cursor mutation seems reasonable, since the mutation sites have the knowledge that they must be in an input state. I try to avoid embedding too many implicit knowledge in an already complex state machinery. It should also lead to less code change than approach (1), but I'll defer to your judgement.

lukhnos (changes requested):
Overall LGTM, but I'd like to raise my concern about making the state machine more complex with the changes this PR introduced.

Source/KeyHandler.mm
```diff
@@ -1834,7 +1853,7 @@ - (void)_walk
     _latestWalk = _grid->walk();
 }
 
-- (InputStateChoosingCandidate *)_buildCandidateState:(InputStateNotEmpty *)currentState useVerticalMode:(BOOL)useVerticalMode
+- (InputStateChoosingCandidate *)_buildCandidateStateFromInputtingState:(InputStateInputting *)inputting   UseVerticalMode:(BOOL)useVerticalMode
```
lukhnos: I believe it should be `useVerticalMode:` instead of `UseVerticalMode:`. Also, there are two extra spaces here.

lukhnos (approved):
Thanks! Just one issue with the method name.

PR #445, opened 2024-03-08
Update associated phrases

lukhnos:
The diff at https://github.com/openvanilla/McBopomofo/commit/825750be3ddc4aeec69bae63c0781bf4fd1d8a91 should give us an easier look at what really changed. PTAL.

The data is generated from the same phrase.occ https://github.com/openvanilla/McBopomofo/commit/969bc7fb3686e1519489b7fb558f2d6b0e839410 using the same script described in https://github.com/openvanilla/openvanilla/pull/79


zonble (approved):
Looks good

PR #451, opened 2024-03-16
Associated Phrases v2

lukhnos:
This brings the macOS version in line with fcitx5-mcbopomofo. Major changes are in the key handler and the input method controller. The code in `Engine/` is just from fcitx5-mcbopomofo.


zonble (approved):
I've tested it. Looks great!

PR #454, opened 2024-03-19
Reset the key label index after scrolling

lukhnos:
If the vertical candidate table view gets scrolled, the key strip label highlight also needs to be readjusted, otherwise the highlighted key label won't match that of the new highlighted row in the table view.

Notice that we only readjust keyLabelStripView.highlightedIndex after the scroll is done. We hide the highlight during the scroll, so as to reduce the visible mismatch between the two highlights.

### DIFF
--- a/Packages/CandidateUI/Sources/CandidateUI/VerticalCandidateController.swift
+++ b/Packages/CandidateUI/Sources/CandidateUI/VerticalCandidateController.swift
@@ -320,7 +320,7 @@ public class VerticalCandidateController: CandidateController {
             self.tableView.scrollRowToVisible(visibleRowIndexes.lowerBound)
 
             let finalSelected = self.selectedCandidateIndex
-            if finalSelected != UInt.max && visibleRowIndexes.contains(Int(finalSelected)) == true {
+            if finalSelected != UInt.max && visibleRowIndexes.contains(Int(finalSelected)) {
                 // Reset the key label strip highlight after the scroll.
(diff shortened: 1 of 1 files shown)


lukhnos:
This fixes an issue that I discovered while reviewing https://github.com/openvanilla/McBopomofo/pull/453; we very much overlooked this when we introduced the delayed scrollRowToVisible logic a few months ago.

Packages/CandidateUI/Sources/CandidateUI/VerticalCandidateController.swift
```diff
@@ -315,18 +307,24 @@ public class VerticalCandidateController: CandidateController {
...
         scrollTimer?.invalidate()
         scrollTimer = Timer.scheduledTimer(withTimeInterval: 0.25, repeats: false) { timer in
-            self.tableView.scrollRowToVisible(visibleRowIndexes.lowerBound)
             self.scrollTimer?.invalidate()
             self.scrollTimer = nil
+
+            self.tableView.scrollRowToVisible(visibleRowIndexes.lowerBound)
+
+            let finalSelected = self.selectedCandidateIndex
+            if finalSelected != UInt.max && visibleRowIndexes.contains(Int(finalSelected)) == true {
```
skyeyvapor: Redundant `== true`.

lukhnos: Thanks for pointing it out! Done.

zonble (approved):
I did not see any issue. Looks good. :)

PR #465, opened 2024-04-07
Updates the Braille converter

zonble:
1. Add proper spaces between Bopomofo Braille syllables, letters and digits.
2. Correctly converts the points and percent symbol


lukhnos (approved):
Thanks!

PR #467, opened 2024-04-11
Update README to bump dev env requirements

lukhnos:
Since the packages BopomofoBraille and ChineseNumbers require Swift 5.9+, we need to update README.md to reflect that.

As it happens, it is also the configuration of our CI actions.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -4,11 +4,9 @@
 
 小麥注音輸入法可以在 macOS 10.13 以上版本運作。如果您要自行編譯小麥注音輸入法，或參與開發，您需要：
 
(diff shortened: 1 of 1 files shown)


RaidenKunikuzushi:
I'm afraid that it's not needed. Both BopomofoBraille and ChineseNumbers are compilable using Swift 5.5.

The only thing need to do is to modify the first line of Package.swift of these two packages from:

```swift
// swift-tools-version: 5.9
```

to:

```swift
// swift-tools-version: 5.5
```

That's enough.

zonble (approved):
Looks good. Thanks!

PR #485, opened 2024-05-22
Fix typo: 磅礡 -> 磅礴

ChiahongHong:
When I was typing the word `磅礴`, I suddenly noticed that it looked weird. I realized that the character was missing the `氵` radical. The correct character should be `礴`, while `礡` is a variant character (異體字). This typo seems to have a long history, as it also appears in the Microsoft Bopomofo, but it is correct in the macOS built-in input method.

---

教育部國語辭典簡編本 https://dict.concised.moe.edu.tw/dictView.jsp?ID=2840

教育部重編國語辭典修訂本 https://dict.revised.moe.edu.tw/dictView.jsp?ID=23998

教育部異體字字典 https://dict.variants.moe.edu.tw/dictView.jsp?ID=65156


xatier:
lgtm, thanks for fixing this!

lukhnos (approved):
Thanks!

PR #499, opened 2024-06-24
Refactor iterator-based loop with structured bindings in Mandarin

ChiahongHong:
This PR improves code readability by replacing the iterator-based loop with [C++17 structured bindings](https://en.cppreference.com/w/cpp/language/structured_binding). This change simplifies the iteration over the map by directly unpacking its elements into variables.


lukhnos (approved):
Thank you! This is indeed much more readable.

PR #515, opened 2024-07-31
Adds 化 and 法 into the heterophony maps

zonble:
McBopomofo has several heterophony maps that helps to fine-tune the frequency of phrases. the PR uses these maps to reduce the frequency of 化 when it is ㄏㄨㄚ, so 花 could be the first.

It also make 法 to be the first so it can be in front of 琺 when it is ㄈㄚˋ.


lukhnos (approved):

xatier:
Would you also consider adding this PR as an example of handling polyphonic characters?

https://github.com/openvanilla/McBopomofo/wiki/%E8%A9%9E%E5%BA%AB%E9%96%8B%E7%99%BC%E8%AA%AA%E6%98%8E

zonble:
加上去了。

PR #522, opened 2024-09-06
Fix phonetics for 波

xatier:
We provide both for fault tolerance.

https://dict.revised.moe.edu.tw/dictView.jsp?ID=44&q=1&word=%E6%B3%A2#order1

### DIFF
--- a/McBopomofoTests/ServiceProviderTests.swift
+++ b/McBopomofoTests/ServiceProviderTests.swift
@@ -34,8 +34,8 @@ final class ServiceProviderTests: XCTestCase {
 
     func testExtractReading1() {
         let provider = ServiceProvider()
-        let output = provider.extractReading(from: "消波塊")
-        XCTAssert(output == "ㄒㄧㄠ-ㄆㄛ-ㄎㄨㄞˋ")
+        let output = provider.extractReading(from: "電腦")
+        XCTAssert(output == "ㄉㄧㄢˋ-ㄋㄠˇ")
     }
 
     func testExtractReading2() {


McBopomofoTests/ServiceProviderTests.swift
```diff
@@ -35,7 +35,7 @@ final class ServiceProviderTests: XCTestCase {
     func testExtractReading1() {
         let provider = ServiceProvider()
         let output = provider.extractReading(from: "消波塊")
-        XCTAssert(output == "ㄒㄧㄠ-ㄆㄛ-ㄎㄨㄞˋ")
```
xatier, line 38: Perhaps we need some tests for polyphonic characters like this?

tianjianjiang: > Perhaps we need some tests for polyphonic characters like this?

I think so.

And when a test fail, it's a perfect timing to realize whether it's a regression. When it's not a regression, it indicates a change of behavior, and a change of behavior better has a good reason and a consensus, IMO.

BTW, a change of behavior can also break backward compatibility by definition, and then it will need a rather significant release in the sense of SemVer.

+CC @lukhnos @mjhsieh @zonble

xatier: I don't consider this is a regression, the code behaves as expected. `extractReading` only returns the first reading; hence we need to update the expected output here to reflect the dictionary update.

That's the reason I was asking for having more tests with polyphonic characters.

tianjianjiang: @xatier That's why I am saying "when it's not a regression..." and I agree it'd be nice to have more tests.

And yet I'd also like to emphasize that, if a test defines the first reading, it should remain the same unless we explicitly want to change the first reading. If not, then it can still be a regression.

The bottom line is, it's usually not a good sign to remove a test, IMO.

lukhnos: Instead of removing the test, how about changing the test string to something that does not have the ambiguity that is introduced in this PR? Something like 電腦 or a phrase that has no alternative (or fault-tolerant) readings.

lukhnos (changes requested):

McBopomofoTests/ServiceProviderTests.swift
xatier: Good idea, updated.

lukhnos (approved):

PR #523, opened 2024-09-07
Fix phonetics for 夫

xatier:
The concise dictionary has the following phrase with `ㄈㄨ˙`.

```text
大夫
工夫
功夫
下功夫
下工夫
頂上功夫
箱底功夫
水磨工夫
一盞茶工夫
```

https://dict.concised.moe.edu.tw/search.jsp?md=1&word=%E5%A4%AB&page=1
https://dict.concised.moe.edu.tw/search.jsp?md=1&word=%E5%A4%AB&page=2
https://dict.concised.moe.edu.tw/search.jsp?md=1&word=%E5%A4%AB&page=3

Also provide both `ㄉㄚˋ ㄈㄨ` and `ㄉㄞˋ ㄈㄨ` for `大夫` for fault tolerance.

https://dict.revised.moe.edu.tw/dictView.jsp?ID=40677&q=1&word=%E5%A4%A7%E5%A4%AB#order1


tianjianjiang (changes requested):
> Also provide both `ㄉㄚˋ ㄈㄨ` and `ㄉㄞˋ ㄈㄨ` for `大夫` for fault tolerance.

Although I'm personally ok with this kind of deviation (even though their meanings are quite different) since there are already wrongly assign syllables, but I recommend not adding `/ㄉㄚˋ ㄈㄨ˙/` because it generally doesn't exist.

xatier:
@tianjianjiang I attempted to cover all combinations for fault tolerance, removed ` ㄉㄚˋ ㄈㄨ˙` as requested.

tianjianjiang:
> @tianjianjiang I attempted to cover all combinations for fault tolerance, removed ` ㄉㄚˋ ㄈㄨ˙` as requested.

@xatier Thank you for that! 🙏

BTW, my preference:
- If it's possible to pronounce in that way, it's ok to add it even if it's not in good references.
- If it's highly unlikely to pronounce in that way, not adding it may keep the syllables informative, e.g., those are actual usages.
  - Even if something like `/ㄉㄚˋ ㄈㄨ˙/` isn't added, if `/ㄈㄨ˙/` is actually of a free morpheme, it will be added as a monosyllable (of 夫 for example), and then "combination" will naturally occur. It may cover reasonable "fault tolerance."

tianjianjiang (approved):

PR #525, opened 2024-09-08
Fix phonetics for 得

xatier:
The concise dictionary has the following phrase with `ㄉㄜ˙`.

```text
得
免得
了得
記得
覺得
曉得
值得
捨得
認得
要得
來得及
靠得住
看得開
看得起
巴不得
不見得
不值得
不由得
了不得
顧不得
怪不得
恨不得
捨不得
說不得
要不得
由不得
美得冒泡
紅得發紫
說得過去
何以見得
忙得不可開交
提得起放得下
```

https://dict.concised.moe.edu.tw/search.jsp?md=1&word=%E5%BE%97&size=-1


tianjianjiang (changes requested):
I think this one is very tricky.

Ideally, `/ㄉㄜ˙/` is for adverbs, and yet MOE dictionary itself isn't consistent. Not to mention that it will create more ambiguity with 的 and 地 (and 底 if it's still in use).

I recommend only adding `/ㄉㄜ˙/` to obvious adverbs.

xatier:
@tianjianjiang this one is indeed tricky, my goal is to cover what concise dictionary provides. Which allows people to type as the phonetics from the dictionary form they learnt from schools.

tianjianjiang (approved):
> to cover what concise dictionary provides. Which allows people to type as the phonetics from the dictionary form they learnt from schools.

@xatier I feel the same for this point, and that's also why I mention the dictionary isn't that consistent. (I should have said dictionaries.)

Anyway, since the PR follows the concise edition, it's totally fine to me. Thank you for the effort!

----
A side note:
Most of non-adverb/non-adjective terms (almost always verbs) don't have `/ㄉㄜ˙/` listed, and yet sometimes the concise edition (for education) and the revised edition (for somewhat linguistic reference) don't agree with each other (e.g., https://dict.revised.moe.edu.tw/dictView.jsp?ID=79619 v. https://dict.concised.moe.edu.tw/dictView.jsp?ID=19116)

xatier:
Indeed, I noticed countless instances that the two dictionaries are not agreeing with each other when auditing the list. In such cases, I usually keep both.

Issue #498, opened 2024-06-21
[功能建議] Allow user-defined phrase frequency


xatier:
**痛點**

I'd like to allow (optional) user-defined phrase frequency in the user's dictionary `data.txt`.

For instance, if I'd like to add this emoji `🎊 ㄍㄨㄥ-ㄒㄧˇ`, I wish to have lower priority than `恭喜`. Currently, the user dictionary ensures user-defined phrases have the highest priority and always shows `🎊` before `恭喜`. Unless the engine has the temporary memory of user selections at the time.

**功能說明**

Allow optional 3rd field in `data.txt`. Keep the current behavior of top priority if this filed is not present.

`🎊 ㄍㄨㄥ-ㄒㄧˇ -8` would make `🎊` to have the same phrase frequency as other emojis and to be on the bottom of the list.
`🎊 ㄍㄨㄥ-ㄒㄧˇ` ensures `🎊` to stay on the top of the list.

In addition to emojis, this also enables users to adjust/override the default frequency (`/usr/share/fcitx5/data/mcbopomofo-data.txt`) of certain words/phrases with their preferences permanently.

Related:

https://github.com/openvanilla/McBopomofo/issues/461
https://github.com/openvanilla/fcitx5-mcbopomofo/issues/144

zonble:
基本上只要把「恭喜」也放進 data.txt，也會提升詞頻。所以只要在 data.txt 裡頭，把「恭喜」放到「🎊」前面就好了。

然後 #649 增加了一種把既有詞彙快速加到 data.txt 的方式。

PR #531, opened 2024-09-26
Fix phonetics for 誼

xatier:
The concise dictionary prefers `ㄧˋ`, we provide both for fault tolerance.

https://dict.concised.moe.edu.tw/search.jsp?md=1&word=%E8%AA%BC#searchL


lukhnos:
This is a problematic entry. I suspect no one reads 交誼 (and its derivative terms such as 交誼廳) as ㄐㄧㄠ ㄧˋ because of our natural tendency to disambiguate very common terms (in this case 交易 which occurs much commonly). I feel that this one borderlines on following the dictionary for the sake of following the dictionary.

xatier:
No worries, feel free to close this one if you don't find it suitable for the dictionary.

Side note, I wonder if we had the capability to support dual/multiple frequency for these case, we could adjust the frequency for `交誼 ㄐㄧㄠ ㄧˋ` to be significantly lower than `交易 ㄐㄧㄠ ㄧˋ`. This could give more flexibility to handle these cases. Similar idea to my proposal in https://github.com/openvanilla/McBopomofo/issues/498 for user dictionary.

xatier:
My planned approach is to keep sharing these PRs when I get some chance, but would like the team to decide whether 小麥 wants to accept the changes or not. We merge only the PRs that the team feels valuable for users. Would that work for you?

lukhnos:
> My planned approach is to keep sharing these PRs when I get some chance, but would like the team to decide whether 小麥 wants to accept the changes or not. We merge only the PRs that the team feels valuable for users. Would that work for you?

It works, and it's much appreciated. I want to reiterate that we're very grateful that we receive PRs like the ones you have contributed.

You [pointed out the inconsistency](https://github.com/openvanilla/McBopomofo/pull/535#issuecomment-2417443017) in McBopomofo's own data. I agree with you. I believe we are all aware of the low-hanging fruits in any attempt to fix something. Many recent PRs in this category had addressed those without any controversy (and that's considering the fact my fellow reviewers have a high bar for those), but we now have some cases where we are reluctant about, and I'm afraid that we have to make a subjective decision on those.

I wish I could be more academic in giving my own rationales but I don't have the resource to make those arguments (such as using corpus and audio sources to show one pronunciation is more prevalent in one variant of Mandarin than the other—that would be the job for MOE's dictionary editors!). There is also the fluid nature of spoken language. 亞 is a good example. While it's very telling which variant you're speaking when you say 亞洲 (this is especially so in formal speech such as news announcments), I'd venture to say the reading for 波希米亞 (third vs fourth tone) is likely equally split among Taiwanese Mandarin speakers. So `波希米亞 ㄅㄛ ㄒㄧ ㄇㄧˇ ㄧㄚˋ` seems a good addition, but `泛亞 ㄈㄢˋ ㄧㄚˋ` is unlikely (it's also not how the now defunct telecom company used to call itself, for example).

All this was to say that there are times when an effort to keep things consistent runs counter to a maintainer's intuition about the language, and unfortunately the rationales will be incomplete or subjective. If I'm doing the review, though, I'll do my best to make a case-by-case argument why I need to make such a subjective call. How does that sound to you?

> Side note, I wonder if we had the capability to support dual/multiple frequency for these case, we could adjust the frequency for 交誼 ㄐㄧㄠ ㄧˋ to be significantly lower than 交易 ㄐㄧㄠ ㄧˋ. This could give more flexibility to handle these cases. Similar idea to my proposal in https://github.com/openvanilla/McBopomofo/issues/498 for user dictionary.

In the case of this PR, I checked the built data and 交易 would still have a higher score, but I agree with you in general that having something as suggested in #498 is helpful.

xatier:
@lukhnos I see we share the same view regarding how we'd like to update the dictionary files here. I'm also relying on my experience and intuition as a native speaker when crafting these PRs. I have no issues if we need to have some subjective calls with certain controversial terms, that's totally fine!

Just kindly let me know and I can help update the PRs. Let's work together to improve the 小麥 dictionary.

xatier:
I have removed the `交誼 ㄐㄧㄠ ㄧˋ` terms, kindly let me know if you seek for other changes. Thanks again for reviewing this and sharing your opinions!

Source/Data/BPMFMappings.txt
```diff
@@ -24738,7 +24738,9 @@
 友盟 ㄧㄡˇ ㄇㄥˊ
 友訊科技 ㄧㄡˇ ㄒㄩㄣˋ ㄎㄜ ㄐㄧˋ
 友誼 ㄧㄡˇ ㄧˊ
+友誼 ㄧㄡˇ ㄧˋ
```
lukhnos: I'd remove 友誼 based on the same argument for 交誼. Otherwise this LGTM.

lukhnos (changes requested):
> I have removed the 交誼 ㄐㄧㄠ ㄧˋ terms, kindly let me know if you seek for other changes. Thanks again for reviewing this and sharing your opinions!

Much appreciated, and thanks for the feedback to my comment. I suggested removing 友誼 but keeping others so as to maintain the PR's purpose as an attempt to follow the official dictionary as much as possible while also maintaining that the changes are subject to our discretion.

Source/Data/BPMFMappings.txt
xatier: sounds good, removed.

lukhnos (approved):
Thanks!

PR #532, opened 2024-09-26
Fix phonetics for 伯

xatier:
`伯伯 ㄅㄛˊ ㄅㄛ˙` is the only one in the concise dictionary that has `ㄅㄛ˙`.

https://dict.concised.moe.edu.tw/search.jsp?md=1&word=%E4%BC%AF#searchL


xatier:
Wondering if we should add `伯伯 ㄅㄟ ㄅㄟ` for slangs, lol

Source/Data/BPMFBase.txt
```diff
@@ -331,6 +331,7 @@
 鸔 ㄅㄛˊ bo2 1i6 big5
 糪 ㄅㄛˊ bo2 1i6 big5
 蔔 ㄅㄛ˙ bo5 1i7 big5
+伯 ㄅㄛ˙ bo5 1i7 big5
```
lukhnos: I don't think we need this entry if we just want to take in 伯伯 as ㄅㄛˊ ㄅㄛ˙, since you can still type ㄅㄛ˙ alone and so it's not an "invalid" syllable.

lukhnos (changes requested):

Source/Data/BPMFBase.txt
xatier: Thanks, this has been removed.

lukhnos (approved):

PR #548, opened 2024-10-22
Fixes the crash that happens while handling the tab key.

zonble:
We used a wrong candidate index when handling the tab key. Fixed.


lukhnos (approved):
LGTM. Thanks!

PR #553, opened 2024-11-02
Revise the associated phrases feature

zonble:
- Now the shift + enter key to insert associated phrases is on by default.
- When the "associated phrase" mode menu item is selected, the associated phrases would appear when each character is entered.
- Also added tests.


McBopomofoTests/AssociatedPhrasesTests.swift
```diff
@@ -0,0 +1,142 @@
...
+import XCTest
+
+@testable import McBopomofo
+
+class AssociatedPhrasesTests: XCTestCase {
+
+    var handler = KeyHandler()
+
+    override func setUpWithError() throws {
+        Preferences.chineseConversionEnabled = false
```
lukhnos, line 33: Question / nit: if a developer runs the test locally, would this disable their Chinese conversion without their knowing?

lukhnos, line 35: Question / nit: if a developer runs the test locally, would this disable their Chinese conversion without their knowing?

lukhnos (approved):
LGTM! Just one question/nit about the test setup.

McBopomofoTests/AssociatedPhrasesTests.swift
zonble: Okay. I Will fix this later.

PR #562, opened 2025-01-03
Add more arrow symbols

xatier:


xatier:
I wonder if we also want to make this file sorted?

```
LC_ALL=c sort -o Symbols.txt Symbols.txt
```

lukhnos:
> I wonder if we also want to make this file sorted?
> 
> ```
> LC_ALL=c sort -o Symbols.txt Symbols.txt
> ```

We can keep it as is for now. I wonder whether sorted entries would make it easier or harder to maintain the file, but at least the symbols would be listed in Unicode order and therefore a reader would be easy to grasp where an entry falls in the Unicode table (emojis, arrows, and so on)? Feel free to create an issue for this.

lukhnos (approved):

xatier:
Yeah, it's a bit tricky as should we group emojis by their orders in the Unicode table, or by the (Chinese translation) semantic meanings? I do not have a good answer for this.

PR #616, opened 2025-02-24
Handle key up events during active input state.

zonble:
This prevents double-space from affecting the current input. While macOS may normally insert a period on double space, this should be suppressed when there is an active composing buffer or candidate window.

- related: #528
- related: #615


lukhnos (approved):
LGTM. Thanks!

PR #620, opened 2025-03-02
chore: initiate a python project structure of data curation scripts

tianjianjiang:
The PR only duplicates Python related files to a typical folder structure. It does not affect any actual data curation process.


lukhnos:
Thanks! It's about time we do some restructuring (including renaming and refactoring) of the Python code in `Data/`. Three concerns with the PR:

- Do we want to call it `ov-data` (with a hyphen in it making it non-trivial to import from in the future)? Could this be `analysis` or `curation`? Or do you expect all future work to live entirely inside `ov-data` and no code will be imported from there? At any rate the prefix `ov-` seems a misnomer since this is not the OpenVanilla project itself…
- Related: should we actually move the Jupyter notebook to the top level so that it can import from the packages? I guess it really depends on how you envision the project and package layout.
- Since this PR duplicates files from `bin/`, I'm concerned that we may step on each other's foot when we refactor, since it's hard to track which files are actually needed for building the database, and which are currently just for analysis/curation purposes.

Thoughts?

tianjianjiang:
> * Do we want to call it `ov-data` (with a hyphen in it making it non-trivial to import from in the future)? Could this be `analysis` or `curation`? Or do you expect all future work to live entirely inside `ov-data` and no code will be imported from there? At any rate the prefix `ov-` seems a misnomer since this is not the OpenVanilla project itself…

Good catch!

Firstly I forgot to change the inner `ov-data` to `ov_data` based on how a Python project usually is: hyphen for package folder (project root; outer `ov-data`) and underscore for source  folder (import path; inner `ov_data`).

And indeed the name itself can be better. Let's use `curation` for now.


> * Related: should we actually move the Jupyter notebook to the top level so that it can import from the packages? I guess it really depends on how you envision the project and package layout.

The import usually links to the inner source folder (`ov_data`) so both `notebooks` and `tests` will import from it. In other words, the `PYTHONPATH` will point to `ov-data/ov_data`.


> * Since this PR duplicates files from `bin/`, I'm concerned that we may step on each other's foot when we refactor, since it's hard to track which files are actually needed for building the database, and which are currently just for analysis/curation purposes.

I was debating with myself whether symlinks could be better. Or perhaps I should have move more files at once based on Makefile and GitHub Actions. For now I've changed those files to symlinks.

The bottom line of my plan is that a file can only be moved after it is fully covered by tests.

lukhnos (changes requested):
The symlinking approach LGTM, but the symlinks don't appear to work. Should they be linked as, for example, `../../bin/cook.py` instead of `Source/Data/bin/cook.py`?

tianjianjiang:
> The symlinking approach LGTM, but the symlinks don't appear to work. Should they be linked as, for example, `../../bin/cook.py` instead of `Source/Data/bin/cook.py`?

@lukhnos I'll try to modify those links again. I didn't want to use symlink because with git this happened to me before.

lukhnos (approved):

lukhnos:
LGTM, thanks!

PR #651, opened 2025-06-21
Prevents applying the boost and exclude keys on the macros and other converted candidates.

zonble:
#649 introduces two new keys to boost and exclude a candidate. However, these keys should not be applied to converted (by Chinese converter or macro converter) candate.

The PR fixes the issue.

### DIFF
--- a/Source/Engine/gramambular2/language_model.h
+++ b/Source/Engine/gramambular2/language_model.h
@@ -45,14 +45,25 @@ class LanguageModel {
   // usually a log probability from a language model.
   class Unigram {
    public:
-    explicit Unigram(std::string val = "", double sc = 0)
-        : value_(std::move(val)), score_(sc) {}
+    explicit Unigram(std::string val = "",
+                     double sc = 0,
+                     std::string originalValue = "",
+                     std::vector<std::string> annotations = std::vector<std::string>()
+                     )
+        : value_(std::move(val)),
+          score_(sc),
+          originalValue_(std::move(originalValue)),
+          annotations_(std::move(annotations)){}
 
     [[nodiscard]] const std::string& value() const { return value_; }
+    [[nodiscard]] const std::string& originalValue() const { return originalValue_; }
+    [[nodiscard]] const std::vector<std::string>& annotations() const { return annotations_; }
     [[nodiscard]] double score() const { return score_; }
 
    private:
     std::string value_;
+    std::string originalValue_;
+    std::vector<std::string> annotations_;
     double score_;
   };
 };
--- a/Source/Engine/gramambular2/language_model.h
+++ b/Source/Engine/gramambular2/language_model.h
@@ -45,25 +45,17 @@ class LanguageModel {
   // usually a log probability from a language model.
   class Unigram {
    public:
-    explicit Unigram(std::string val = "",
-                     double sc = 0,
-                     std::string originalValue = "",
-                     std::vector<std::string> annotations = std::vector<std::string>()
-                     )
-        : value_(std::move(val)),
-          score_(sc),
-          originalValue_(std::move(originalValue)),
-          annotations_(std::move(annotations)){}
+    explicit Unigram(std::string val = "", double sc = 0,
+                     std::string rawValue = "")
+        : value_(std::move(val)), score_(sc), rawValue_(std::move(rawValue)) {}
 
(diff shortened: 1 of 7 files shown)


Source/KeyHandler.mm
```diff
@@ -1392,39 +1402,11 @@ - (BOOL)_handleCandidateState:(InputState *)state
...
-                return YES;
-            }
-            __weak __typeof__(self) weakSelf = self;
-            NSArray *entries = @[
-                [[InputStateCustomMenuEntry alloc] initWithTitle:NSLocalizedString(@"Exclude",
+                }];
+                [entries addObject:boost];
+                title =[NSString stringWithFormat:NSLocalizedString(@"Do you want to boost the score of the phrase \"%@\"?", @""), candidate.value];
+            } else if (isMinusKey) {
+                InputStateCustomMenuEntry *exlude = [[InputStateCustomMenuEntry alloc] initWithTitle:NSLocalizedString(@"Exclude",
```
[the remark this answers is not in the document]
zonble: Fixed

Source/Engine/gramambular2/language_model.h
```diff
@@ -45,14 +45,25 @@ class LanguageModel {
...
+          annotations_(std::move(annotations)){}
 
     [[nodiscard]] const std::string& value() const { return value_; }
+    [[nodiscard]] const std::string& originalValue() const { return originalValue_; }
+    [[nodiscard]] const std::vector<std::string>& annotations() const { return annotations_; }
     [[nodiscard]] double score() const { return score_; }
 
    private:
     std::string value_;
+    std::string originalValue_;
```
lukhnos: I would call this `rawValue_` and rename all references to `originalValue` to `rawValue`, and add a comment on the getter above that `rawValue()` returns the value _as retrieved from a unigram language model_, where as the actual `value()` may be a transformed or converted value such as those by a macro converter.

Source/Engine/gramambular2/language_model.h
```diff
@@ -45,14 +45,25 @@ class LanguageModel {
...
+    std::vector<std::string> annotations_;
```
lukhnos: I don't see `annotations()` be used anywhere. Do we really need this information that causes every `Unigram` constructor to have an necessary `std::vector` construction and move?

Source/KeyHandler.mm
```diff
@@ -1364,12 +1364,19 @@ - (BOOL)_handleCandidateState:(InputState *)state
...
+    if (_inputMode == InputModeBopomofo && [state isKindOfClass:[InputStateChoosingCandidate class]]) {
+        BOOL isPlusKey = [input.inputText isEqualToString:@"="] || [input.inputText isEqualToString:@"+"];
+        BOOL isMinusKey = [input.inputText isEqualToString:@"-"] || [input.inputText isEqualToString:@"_"];
+        if (isPlusKey || isMinusKey) {
             InputStateChoosingCandidate *currentState = (InputStateChoosingCandidate *)state;
             NSInteger index = gCurrentCandidateController.selectedCandidateIndex;
             InputStateCandidate *candidate = currentState.candidates[index];
             NSString *reading = candidate.reading;
+
+            if (![candidate.value isEqualToString:candidate.originalValue]) {
```
lukhnos: This needs a comment, otherwise it's hard to understand why we bail on this condition. Also see my comment on renaming `originalValue` to `rawValue`.

lukhnos (approved):
LGTM, but I left some comments regarding field naming, adding comments, and whether the `annotations_` field is really necessary. Thanks!

Source/Engine/gramambular2/reading_grid.h
```diff
@@ -183,10 +183,11 @@ class ReadingGrid {
   WalkResult walk();
 
   struct Candidate {
-    Candidate(std::string r, std::string v)
-        : reading(std::move(r)), value(std::move(v)) {}
+    Candidate(std::string r, std::string v, std::string ov = "")
+        : reading(std::move(r)), value(std::move(v)), originalValue(std::move(ov)) {}
     const std::string reading;
     const std::string value;
+    const std::string originalValue;
```
lukhnos: Forgot to add that I'd also rename this to `rawValue`.

Source/KeyHandler.mm
zonble: Resolved.

PR #664, opened 2025-09-06
Use correct data type for SQLite.swift

zonble:
The SQLite.swift library has updated its API to namespace the `Expression` type under `SQLite`. This commit updates the code to use `SQLite.Expression` to ensure compatibility with the latest version of the library.

This change is necessary to build the project with the updated dependency and fixes potential compilation errors.


lukhnos (approved):
LGTM. Thanks!

PR #715, opened 2025-10-23
Update CI workflow to package and upload artifact

ChiahongHong:
In #688 and #478, users who are not familiar with building from source may have to wait until the next McBopomofo release to get the issues fixed.

Since our CI runs a build for every push, packaging and uploading the build as an artifact could allow users to download it directly and test whether the reported issues have been resolved.

By the way, I noticed that even though it’s built with the release scheme, the unzipped McBopomofo-Installer is larger than the official release. I’m so curious what optimizations were applied in the official version XD

### Demo
https://github.com/ChiahongHong/McBopomofo/actions/runs/18755941747

### Known Issue
Because the app is not code-signed, users will need to manually allow it in System Settings → Privacy & Security.

### Version Number
For now, I’m using the short commit SHA as `CFBundleVersion`. I would make changes if you have any suggestions.

### DIFF
--- a/.github/workflows/continuous-integration-workflow-xcode-latest.yml
+++ b/.github/workflows/continuous-integration-workflow-xcode-latest.yml
@@ -12,6 +12,13 @@ jobs:
       - uses: maxim-lobanov/setup-xcode@v1
         with:
           xcode-version: "^16.1.0"
+      - name: Get commit SHA
+        run: echo "SHA=$(git rev-parse --short HEAD)" >> $GITHUB_ENV
+      - name: Update CFBundleVersion with commit SHA
+        run: |
+          echo "Updating CFBundleVersion to ${SHA}"
+          /usr/libexec/PlistBuddy -c "Set :CFBundleVersion ${SHA}" "Source/Installer/Installer-Info.plist"
+          /usr/libexec/PlistBuddy -c "Set :CFBundleVersion ${SHA}" "Source/McBopomofo-Info.plist"
       - name: Build McBopomofoLMLibTest
         run: cmake -DENABLE_TEST=1 -S . -B build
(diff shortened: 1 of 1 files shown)


lukhnos:
> Because the app is not code-signed, users will need to manually allow it in System Settings → Privacy & Security.

Thanks for the PR. I also saw [your comment](https://github.com/openvanilla/McBopomofo/issues/688#issuecomment-3436685992) proposing we provide the built artifact (the installer) along with CI actions.

Let me discuss this with the other maintainers. I personally don't prefer that we developers encourage anyone to override macOS security restrictions in order to run an unsigned binary. Especially for something that interacts so intimately with user inputs.

If we really want to do something like nightly builds, they will have to be signed. Telling someone who's really enthusiastic with fresh new features to check out the source code and build with Xcode themselves is a high bar and I recognize the technical hurdles, but it's a better (or at least less bad) path compared to the practice of overriding security restrictions.

> By the way, I noticed that even though it’s built with the release scheme, the unzipped McBopomofo-Installer is larger than the official release. I’m so curious what optimizations were applied in the official version XD

The installer built with the release scheme carries the built McBopomofo.app uncompressed, whereas the official installer comes with the compressed McBopomofo.app, hence the bundle size difference. :)

ChiahongHong:
> Let me discuss this with the other maintainers. I personally don't prefer that we developers encourage anyone to override macOS security restrictions in order to run an unsigned binary. Especially for something that interacts so intimately with user inputs.

That makes sense to me. Feel free to close this PR.

> The installer built with the release scheme carries the built McBopomofo.app uncompressed

Oh, I didn’t realize it was that :rofl: thanks for the explanation!

lukhnos:
Thanks! Closing this now.

PR #729, opened 2025-11-05
Fix phonetics for 解

xatier:
The concise dictionary prefers `解送 ㄐㄧㄝˋ ㄙㄨㄥˋ` and `押解 ㄧㄚ ㄐㄧㄝˋ`

Ref: https://dict.concised.moe.edu.tw/search.jsp?md=1&word=%E8%A7%A3#searchL


Source/Data/heterophony2.list
```diff
@@ -53,6 +53,7 @@
 處 ㄔㄨˇ
 行 ㄏㄤˊ
 要 ㄧㄠ
+解 ㄒㄧㄝˋ
```
lukhnos: You can revert this, `解 ㄐㄧㄝˇ`  is already in `heterophony1.list`, which pushes the scores of the other two readings (`解 ㄒㄧㄝˋ` and `解 ㄐㄧㄝˋ`) down. Putting those in `heterophony{2,3}.list` actually promotes their score, which is what we intend with this PR.

Source/Data/heterophony3.list
```diff
@@ -11,5 +11,6 @@
 著 ㄓㄠˊ
 著 ㄓㄨˋ
 著 ㄓㄨㄛˊ
+解 ㄐㄧㄝˋ
```
lukhnos: Ditto. Let's revert the commit that introduces the change in the heterophony files.

lukhnos (changes requested):

Source/Data/heterophony2.list
xatier: done.

lukhnos (approved):
LGTM. Thanks!

PR #730, opened 2025-11-05
Fix phonetics for 拾

xatier:
The concise dictionary prefers `拾 ㄕㄜˋ`

Ref: https://dict.concised.moe.edu.tw/search.jsp?md=1&word=%E6%8B%BE#searchL


Source/Data/heterophony2.list
```diff
@@ -29,6 +29,7 @@
 尋 ㄒㄧㄣˊ
 屯 ㄓㄨㄣ
 待 ㄉㄞ
+拾 ㄕㄜˋ
```
lukhnos: Again let's revert the commit with this change. This actually promotes the score of `捨 ㄕㄜˋ`. Although it's not as high as interfering with other more common ㄕㄜˋ charcaters, this promotion is also not really what this PR intends to achieve.

lukhnos (changes requested):

lukhnos (approved):
LGTM.

PR #750, opened 2025-11-23
Fix SyntaxWarning in regex using raw string

ChiahongHong:
When running with `python -m`, Python only shows a given warning once, so the message disappears after the first run unless `__pycache__` is removed.

That’s probably why we missed it before. Still, it makes sense to clean this up : )


lukhnos (approved):
LGTM. Thanks!

PR #768, opened 2026-01-05
Bump version to 2.9.5 (2241)

lukhnos:


lukhnos:
We'll need to bump version after the recently discovered regression is fixed.

PR #810, opened 2026-04-03
Further fine-tune Gemini code review settings

lukhnos:
* Ignore database source since Gemini only tells us "unable to generate a review [...] due to file types involved not being currently supported."
* Only generate a maximum of 5 critical issues to focus.


zonble (approved):
Looks good. :)

PR #817, opened 2026-04-12
Add  full width letters to the symbol list

zonble:
The PR add alternative letters such as Ａ, Ⓐ, 🅐. To input these letters, press the grave key (`) and then type A. The keys are including A to Z.


lukhnos (approved):
LGTM.

PR #824, opened 2026-05-02
Reduce the amount of candidates in the auto trggered associated phrases in the smart mode

zonble:
We have implemented a feature to show associated phrases in smart Bopomofo mode. However, the feature is quite annoying since it always gives too many candidates.

The PR reduces the candidates. Not it only shows one candidate and a user can use Shift+Enter to commit it to the composing buffer. If a user wants more candidates, he or she can use the Tab key to show the full candidate list.

image


lukhnos (approved):
LGTM. Thanks!

PR #837, opened 2026-05-30
Fix data checker by forcing LC_ALL=C

ChiahongHong:
This PR fixes our data checker, which has apparently been broken for quite a while...

When `LC_ALL` is not explicitly set, the default locale is usually something like `en_US.UTF-8`. Under that locale, `sort -u` may collapse many CJK strings into the same collation group, which means `make check` under `Data` was basically failing to catch real differences : )

Since #833 added new entries without adding them to `phrase.occ`, causing the check to fail. I discovered this while preparing another PR, because GitHub Actions started failing on the `macos-15-arm64 20260527.0100.1` runner image. It may have changed some locale settings or tool versions.

```
$ make check
(empty)
```

```
$ LC_ALL=C make check

--- /dev/fd/63  2026-05-31 02:55:52
+++ /dev/fd/62  2026-05-31 02:55:52
@@ -1555,7 +1555,6 @@
 丁字形
 丁字褲
 丁字鎬
-丁學忠
 丁守中
 丁憂
 丁母憂
@@ -9871,7 +9870,6 @@
 何月何日
 何樂不為
 何樂而不為
-何欣純
 何止
 何況
 何況是
 ......
```


lukhnos (approved):
LGTM. Thanks!

PR #777, opened 2026-01-28
Refactor DAG shortest path algorithm using Viterbi

ChiahongHong:
Recently, I’ve been reading the "NLP bible," [**_Speech and Language Processing_**](https://web.stanford.edu/~jurafsky/slp3/), which explains that for HMMs, the most standard and efficient algorithm is Viterbi.

[Hidden Markov Models - https://web.stanford.edu/~jurafsky/slp3/A.pdf](https://web.stanford.edu/~jurafsky/slp3/A.pdf)

## Viterbi

The Viterbi algorithm is a DP approach used to find the most likely sequence of hidden states

↑↑↑ This looks a bit complex, so let’s consider a simpler version instead
- https://vene.ro/mlsd/lec08_dp.pdf
- This is the pseudocode I actually followed for the implementation XDD

- Forward Pass: maximum score for each position `i` by relaxing all possible spans ending at that position
- Backward Pass: reconstruct the optimal path by backtracking from the end of the grid to the start using the stored back-pointers
- Time Complexity: Maintained $O(V+E)$

## Topological Sort Removed

Also, I realized that `ReadingGrid` itself is a linear lattice, so there is no need to perform a topological sort. It is already inherently topologically ordered:
- Every node starts at index `i` and ends at `i + spanLen`
- Since `spanLen >= 1`, an edge always points to a higher index (strictly forward)
- Therefore, the array indices  already constitute a valid topological order. By iterating linearly through the indices, we satisfy the topological requirement implicitly

## Performance

- MacBook Air M2
- Because the nature of the algorithm is different, the meanings of _vertices_ and _edges_ change accordingly
  - vertices are the reachable states
  - edges are the candidate word transitions 

### Before

- Stress test elapsed: 8069 microseconds

```
[==========] Running 21 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 21 tests from ReadingGridTest
[ RUN      ] ReadingGridTest.Span
[       OK ] ReadingGridTest.Span (32 ms)
[ RUN      ] ReadingGridTest.ScoreRankedLanguageModel
[       OK ] ReadingGridTest.ScoreRankedLanguageModel (0 ms)
[ RUN      ] ReadingGridTest.BasicOperations
[       OK ] ReadingGridTest.BasicOperations (0 ms)
[ RUN      ] ReadingGridTest.InvalidOperations
[       OK ] ReadingGridTest.InvalidOperations (0 ms)
[ RUN      ] ReadingGridTest.DeleteAfterCursor
[...]

PR #838, opened 2026-05-30
Avoid shared_ptr copies in Span::nodeOf

ChiahongHong:
This PR is a follow-up to #777. After that refactor, I kept looking into the remaining hot path and found that `Span::nodeOf()` was still returning `NodePtr` by value. Since `NodePtr` is a `std::shared_ptr<Node>`, every call to `nodeOf()` copied a `shared_ptr`, causing unnecessary atomic reference count updates (which are very expensive) while walking the grid.

This PR changes `Span::nodeOf()` to return `const NodePtr&` instead. The method only returns an existing entry from the span's internal node array, so there is no need to create a new `shared_ptr` owner each time.

This is a small API-level change, but it removes a surprisingly expensive operation from the hottest loop XDD 

## Performance

- MacBook Air M2
- Release mode
- `runGramambular2Test` stress test

### Before

| Round 1 | Round 2 | Round 3 |
| ------- | ------- | ------- |
| 199 us  | 209 us  | 203 us  |

### After

| Round 1 | Round 2 | Round 3 |
| ------- | ------- | ------- |
| 132 us  | 145 us  | 125 us  |

### DIFF
--- a/Source/Engine/gramambular2/reading_grid.cpp
+++ b/Source/Engine/gramambular2/reading_grid.cpp
@@ -577,7 +577,7 @@ void ReadingGrid::Span::removeNodesOfOrLongerThan(size_t length) {
   }
 }
 
-ReadingGrid::NodePtr ReadingGrid::Span::nodeOf(size_t length) const {
+const ReadingGrid::NodePtr& ReadingGrid::Span::nodeOf(size_t length) const {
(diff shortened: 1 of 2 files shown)


ChiahongHong:
For the macOS 15 CI test failure, please refer to #837

lukhnos (approved):
LGTM! This is such a great bang for the buck.

PR #839, opened 2026-05-31
Add missing entries to phrase.occ

lukhnos:
- "崐" from https://github.com/openvanilla/McBopomofo/pull/833
- "𡻈" from https://github.com/openvanilla/McBopomofo/pull/75

Those are the only two characters with score -99.0. For 𡻈 it's problematic because it's ranked lower than the emoji 🪡 because of it.

Also fixes the `sort` Makefile recipe based on how the issue was fixed in https://github.com/openvanilla/McBopomofo/pull/837


lukhnos:
@ChiahongHong Could you help take a pass at this? Thank you!

Source/Data/Makefile
```diff
@@ -64,7 +66,7 @@ check:
 	@awk 'length($$1)/3!=NF-1' BPMFMappings.txt
 	@awk '{print $$2,$$3,$$4,$$5,$$6,$$7}' BPMFMappings.txt | awk '/一/'
 	@diff -u <(awk '{print $$1}' BPMFMappings.txt|sort -u) \
-                 <(awk 'length($$1)>3{print $$1}' phrase.occ|sort -u)
```
lukhnos, line 67: This line is used to remove single-character entries in `phrase.occ` before comparing it to `BPMFMappings.txt`, and here is where the "each character is 3 bytes long" assumption broke down after this PR's change…

ChiahongHong: @lukhnos I was just about to reply, but I was one step too late XD. It does look like this is indeed an issue with CJK Unified Ideographs Extension B.

lukhnos: Now the build is back to green, I think we're all good. Thanks for taking a look!

zonble (approved):

PR #841, opened 2026-06-02
Fix moved-from MemoryMappedFile length to 0

ChiahongHong:
This PR is a very small fix for `MemoryMappedFile` move construction.

Currently, the move constructor resets the moved-from `length_` to -1. Since `length_` is a `size_t`, that value becomes `0xffffffffffffffff` instead of a sensible empty length.

This does not affect current runtime behavior, because only tests move-construct `MemoryMappedFile` objects. The runtime code uses move assignment instead. Still, resetting the moved-from length to 0 keeps the object in a cleaner and more consistent state, and avoids surprising behavior in future code.

### DIFF
--- a/Source/Engine/MemoryMappedFile.cpp
+++ b/Source/Engine/MemoryMappedFile.cpp
@@ -49,7 +49,7 @@ MemoryMappedFile& MemoryMappedFile::operator=(
 MemoryMappedFile::~MemoryMappedFile() { close(); }
 
 bool MemoryMappedFile::open(const char* path) {
-  if (data_) {
+  if (fd_ != -1) {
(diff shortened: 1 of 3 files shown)


xatier:
Looks good to me, it's surely better to have values set consistently.

lukhnos (changes requested):
LGTM. Thanks! Would it be too much of trouble to also pick up the changes in https://github.com/openvanilla/fcitx5-mcbopomofo/pull/258?

ChiahongHong:
Done～

lukhnos (approved):
LGTM. Thanks!

PR #844, opened 2026-06-10
refactor(engine): derive span length from language model maxKeyLength

tianjianjiang:
## Summary

Replaces the refactor portion of #779, per review feedback there: instead of extracting the walk into a strategy interface, this PR makes the **smallest possible standalone change** — span length becomes dynamic, driven by the language model.

- `Span` storage: fixed `std::array<NodePtr, 8>` → `std::vector` grown on demand
- New `LanguageModel::maxKeyLength()`; returning 0 (the default) keeps today's limit of 8, now named `kDefaultMaxSpanLength`
- `ReadingGrid` derives `maxSpanLength()` from the model at construction; grid window math uses it
- `Span::nodeOf()` keeps the const-reference return from #841 (no shared_ptr copies); out-of-range lookups return a null sentinel instead of asserting

The walk algorithm is **untouched** — no strategy abstraction, no behavioral change to the forward-pass DP from #777.

## Why

This is the enabling refactor for a contextual user model (follow-up PRs): user phrases can exceed 8 syllables, and the grid needs to accept nodes of that length when the model supports them. It is split out so it can be reviewed and verified independently.

## Verification

- All 21 baseline tests pass unchanged; new `DynamicSpanLength` test covers a 10-syllable model vs the default
- Perf guard (8001-reading `StressTest`, 3 runs each): master median **1580µs**, this branch **1499µs** — no regression, identical vertex/edge counts (8001/16001)
- No new files, so no CMake/pbxproj changes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Source/Engine/gramambular2/language_model.h
+++ b/Source/Engine/gramambular2/language_model.h
@@ -41,6 +41,10 @@ class LanguageModel {
   virtual std::vector<Unigram> getUnigrams(const std::string& reading) = 0;
   virtual bool hasUnigrams(const std::string& reading) = 0;
 
+  // The maximum reading key length, in syllables, that this model can match.
+  // Returns 0 if unknown, in which case the grid uses its default span length.
+  virtual size_t maxKeyLength() const { return 0; }
+
   // An immutable unigram with an actual value, along with a score, which is
   // usually a log probability from a language model.
   class Unigram {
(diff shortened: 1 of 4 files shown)


lukhnos:
See #779. Also there have been quite a few PRs on `ReadingGrid` of late, and any refactoring PRs need to be prepared to catch up with those.

PR #853, opened 2026-06-24
Add new entries for '𰻝', '𰻞', and '𰻞𰻞麵' in BPMFBase.txt and phrase.occ

zonble:
See https://zh.wikipedia.org/zh-tw/%F0%B0%BB%9E%F0%B0%BB%9E%E9%BA%B5


lukhnos (approved):
LGTM.

lukhnos (changes requested):
Looking at `Makefile` again, I wonder if some changes are warranted. See my comment.

Source/Data/Makefile
```diff
@@ -61,12 +61,13 @@ sort:
...
 
 check: SHELL:=/bin/bash
-check: export LC_ALL=C
+check: export LC_ALL=en_US.UTF-8
 check:
-	@awk 'length($$1)/3!=NF-1' BPMFMappings.txt
-	@awk '{print $$2,$$3,$$4,$$5,$$6,$$7}' BPMFMappings.txt | awk '/一/'
-	@diff -u <(awk '{print $$1}' BPMFMappings.txt|sort -u) \
-                 <(awk 'length($$1)>4{print $$1}' phrase.occ|sort -u)
+# 	@brew install gawk
```
lukhnos: Remove this, and check if gawk is installed, and fail if not + prompt the user to install it? This should at least make a less frustrating dev experience...

Source/Data/Makefile
```diff
@@ -61,12 +61,13 @@ sort:
...
+	@gawk 'length($$1)!=NF-1' BPMFMappings.txt
+	@gawk '{print $$2,$$3,$$4,$$5,$$6,$$7}' BPMFMappings.txt | gawk '/一/'
+	@diff -u <(gawk 'length($$1)>1{print $$1}' BPMFMappings.txt|sort -u) \
```
lukhnos: Should we introduce a quick Python code (or a Perl one-liner) that _actually_ filters phrases with more than one true Han character instead of relying on an external dependency that is `gawk`? It's not installed on macOS by default, and I still prefer that everything works out of the box with just your own Xcode installation...

lukhnos (changes requested):
Left a comment re the use of `gawk`.

Source/Data/BPMFMappings.txt
```diff
@@ -144962,3 +144962,6 @@
 龜鑑 ㄍㄨㄟ ㄐㄧㄢˋ
 龜頭 ㄍㄨㄟ ㄊㄡˊ
 龜鱉目 ㄍㄨㄟ ㄅㄧㄝ ㄇㄨˋ
+𰻝 ㄅㄧㄤˊ
+𰻞 ㄅㄧㄤˊ
```
ChiahongHong: I think `𰻝 ㄅㄧㄤˊ` and `𰻞 ㄅㄧㄤˊ` should be removed from `BPMFMappings.txt`, since `BPMFMappings` should only contain phrases, not single characters.

zonble: Fixed in f3c63ca93cfb290bf7f66211a27fe772eb1c5ee9

Source/Data/Makefile
zonble: Rev 2166bd1eabe7bbdfab34d25db6b5878b03f4a154 adds a Python script for that.

Source/Data/Makefile
zonble: Fixed.

Source/Data/Makefile
```diff
@@ -63,10 +63,7 @@ sort:
 check: SHELL:=/bin/bash
 check: export LC_ALL=C
 check:
-	@awk 'length($$1)/3!=NF-1' BPMFMappings.txt
-	@awk '{print $$2,$$3,$$4,$$5,$$6,$$7}' BPMFMappings.txt | awk '/一/'
-	@diff -u <(awk '{print $$1}' BPMFMappings.txt|sort -u) \
-                 <(awk 'length($$1)>4{print $$1}' phrase.occ|sort -u)
+	@python3 scripts/check.py
```
lukhnos, line 66: But if `check.py` finds anything incorrect, does it end the program with a non-0 exit code?

lukhnos (changes requested):
Overall LGTM, but `check.py` doesn't seem to exit with non-0 if it finds issues?

lukhnos (approved):
Thanks!

PR #854, opened 2026-06-24
Create simple apps that only contains the candidate window and the tooltip to help UI development

zonble:
The PR add two simple Mac App build target. One app has only one simple candidate window while another has only one tooltip window.

The PR helps developers to apply new UI design easily. When a developer modifies the UI, he or she can just run these simple apps, but not to kill the input method application and run it again.

### DIFF
--- a/Packages/CandidateUI/Package.swift
+++ b/Packages/CandidateUI/Package.swift
@@ -8,7 +8,11 @@ let package = Package(
     products: [
         .library(
             name: "CandidateUI",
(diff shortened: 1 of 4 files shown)


lukhnos (approved):
LGTM. Want to take a look at the Gemini review comments?

zonble:
> LGTM. Want to take a look at the Gemini review comments?

Actually the warnings do not matter. We can use arbitrary text to do preview.

PR #858, opened 2026-06-28
feat(input): prefer exact phrase matches for full input

skyeyvapor:
## Summary

This PR adds an optional composition preference that prefers an exact phrase match when the full input exactly matches a known phrase.

When enabled, if the entire composing buffer is a known phrase, the engine presents that phrase before a higher-scoring segmentation made from shorter words. This only applies to the full input and does not boost the same phrase when it is merely a prefix of a longer input.

## Motivation

Some known phrases can currently lose to a segmentation made from high-frequency single-character words, even when the user has entered the full reading of the intended phrase.

Examples:

| Input | Intended phrase | Previous result | Why this is surprising |
|---|---|---|---|
| ㄗˋ ㄏㄨㄟˋ | 字彙 | 自會 | The full input matches 字彙, but higher-frequency single-character words can be selected instead. |
| ㄧ ㄗㄞˋ | 一再 | 一在 | The full input matches 一再, but it can be split into two single-character words. |
| ㄕˋ ㄕˋ | 試試 | 是是 | The high-frequency 是 can outweigh the known phrase 試試. |
| ㄍㄜˋ ㄕˋ | 各式 | 個是 | Common single-character words can produce a higher score despite an unnatural phrase. |
| ㄉㄜ˙ ㄔㄨ | 得出 | 的出 | The high-frequency 的 can cause an unintended split. |
| ㄧ ㄐㄧㄡˋ | 依舊 | 一就 | The known phrase exists, but the split result can score higher. |
| ㄓㄨㄥ ㄧ | 中醫 | 中一 | A complete domain phrase can lose to a split result. |
| ㄅㄨˋ ㄧˇ | 不已 | 不以 | The phrase has a clear meaning, but the split can rank higher. |
| ㄅㄨˋ ㄉㄠˋ | 步道 | 不到 | If the user intends 步道, the previous behavior can conflict with that intent. |
| ㄕˋ ㄧㄝˇ | 視野 | 是也 | High-frequency single-character words can outweigh the phrase. |
| ㄧㄡˇ ㄕˋ | 有事 | 有是 | A natural phrase can lose to scoring from shorter words. |
| ㄇㄧㄢˋ ㄕˋ | 面試 | 面是 | A common phrase can lose because 是 is scored as a high-frequency single-character word. |
| ㄗㄨㄛˋ ㄕˋ | 做事 | 作是 | A natural verb-object phrase can be split into less natural single-character words. |
| ㄏㄠˇ ㄕˋ | 好事 | 好是 | The intended phrase is common, but the split result may still score higher. |
| ㄕˋ ㄏㄡˋ | 事後 | 是後 | The phrase has a clear meaning, but the split can rank higher. |
| ㄒㄧㄣ ㄕˋ | 心事 | 心是 | A natural phrase can be split into a less natural result. |
| ㄧㄠˋ ㄕ˙ | 鑰匙 | 要時 | A known everyday word can lose to a higher-scoring split result. |

## Behavior

This preference is disabled by default.

When enabled:

- Full-input exact phrase matches are preferred.
- User overrides still take precedence.
- Prefix matches inside longer input are not boosted.
- Normal lattice scoring continues to apply when there is no full-input phrase match.

## Notes

This is intended to handle cases where the user's full input strongly indicates a known phrase, while keeping the existing scoring behavior as the default.

The exact-match shortcut now checks for overridden nodes in the grid before returning a full-input phrase. If any override exists, the engine falls back to the normal Viterbi flow instead of replacing the result afterward. This avoids bypassing user-selected or user-override-model-selected candidates.

When the exact-match preference is enabled, user override suggestions are checked against both paths:

- The exact-match path is checked first, preserving same-reading phrase replacements, such as 字彙 -> 自會.
- If no suggestion is found, the normal Viterbi path is checked as a fallback, preserving corrections from split words to phrases, such as 好顯 -> 好險.

Suggestions found from the exact-match path are applied as high-score overrides so they remain selected after the override disables the exact-match shortcut for the final walk.

### DIFF
--- a/McBopomofoTests/PreferencesTests.swift
+++ b/McBopomofoTests/PreferencesTests.swift
@@ -193,11 +193,11 @@ final class PreferencesTests {
         #expect(Preferences.phraseReplacementEnabled == true)
     }
 
-    @Test("Test standalone phrase boundary toggle")
-    func testStandalonePhraseBoundaryEnabledKey() {
-        #expect(Preferences.standalonePhraseBoundaryEnabled == false)
-        Preferences.standalonePhraseBoundaryEnabled = true
-        #expect(Preferences.standalonePhraseBoundaryEnabled == true)
+    @Test("Test prefer exact phrase match for full input toggle")
+    func testPreferExactPhraseMatchForFullInputKey() {
+        #expect(Preferences.preferExactPhraseMatchForFullInput == false)
+        Preferences.preferExactPhraseMatchForFullInput = true
+        #expect(Preferences.preferExactPhraseMatchForFullInput == true)
     }
 
     @Test("Test Chinese conversion style setting")
--- a/McBopomofoTests/PreferencesTests.swift
+++ b/McBopomofoTests/PreferencesTests.swift
@@ -193,6 +193,13 @@ final class PreferencesTests {
         #expect(Preferences.phraseReplacementEnabled == true)
     }
 
+    @Test("Test standalone phrase boundary toggle")
+    func testStandalonePhraseBoundaryEnabledKey() {
+        #expect(Preferences.standalonePhraseBoundaryEnabled == false)
+        Preferences.standalonePhraseBoundaryEnabled = true
+        #expect(Preferences.standalonePhraseBoundaryEnabled == true)
+    }
+
     @Test("Test Chinese conversion style setting")
     func testChineseConversionStyle() {
         #expect(Preferences.chineseConversionStyle == .output)
--- a/McBopomofoTests/ServiceProviderTests.swift
+++ b/McBopomofoTests/ServiceProviderTests.swift
@@ -148,6 +148,26 @@ final class ServiceProviderTests {
         #expect(output == expected, "\(output)")
     }
 
+    @Test("Test Unicode Taiwanese Braille service respects exact phrase preference")
+    func testConvertUnicodeBrailleToChineseRespectsExactPhrasePreference() {
+        LanguageModelManager.loadDataModels()
+        let previousValue = Preferences.preferExactPhraseMatchForFullInput
+        defer {
+            Preferences.preferExactPhraseMatchForFullInput = previousValue
+        }
+
+        let provider = ServiceProvider()
+        let helper = ServiceProviderInputHelper()
+        provider.delegate = helper as? any ServiceProviderDelegate
+        let input = provider.convertToUnicodeBraille(string: "字彙")
+
+        Preferences.preferExactPhraseMatchForFullInput = false
+        #expect(provider.convertUnicodeBrailleToChineseText(string: input) == "自會")
+
+        Preferences.preferExactPhraseMatchForFullInput = true
+        #expect(provider.convertUnicodeBrailleToChineseText(string: input) == "字彙")
+    }
+
     @Test(
         "Test coverting Chinese to Unicode Taiwanese Braille, then coverting it back",
         arguments: [
--- a/Source/Base.lproj/preferences.xib
+++ b/Source/Base.lproj/preferences.xib
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="24506" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES">
+<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="24765" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES">
     <dependencies>
         <deployment identifier="macosx"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="24506"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="24765"/>
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
     </dependencies>
     <objects>
@@ -28,7 +28,7 @@
             <windowStyleMask key="styleMask" titled="YES" closable="YES" miniaturizable="YES"/>
             <windowPositionMask key="initialPositionMask" leftStrut="YES" rightStrut="YES" topStrut="YES" bottomStrut="YES"/>
             <rect key="contentRect" x="401" y="295" width="475" height="567"/>
-            <rect key="screenRect" x="0.0" y="0.0" width="1920" height="1050"/>
+            <rect key="screenRect" x="0.0" y="0.0" width="1800" height="1129"/>
             <view key="contentView" id="2">
                 <rect key="frame" x="0.0" y="0.0" width="475" height="567"/>
                 <autoresizingMask key="autoresizingMask"/>
@@ -108,7 +108,7 @@
                     </textFieldCell>
                 </textField>
                 <popUpButton verticalHuggingPriority="750" translatesAutoresizingMaskIntoConstraints="NO" id="bhe-DC-FkH">
-                    <rect key="frame" x="232" y="526" width="127" height="24"/>
+                    <rect key="frame" x="232" y="526" width="126" height="24"/>
                     <popUpButtonCell key="cell" type="push" title="Standard" bezelStyle="rounded" alignment="left" lineBreakMode="truncatingTail" state="on" borderStyle="borderAndBezel" imageScaling="proportionallyDown" inset="2" selectedItem="PTu-Ks-u1l" id="tPE-G4-rho">
                         <behavior key="behavior" lightByBackground="YES" lightByGray="YES"/>
                         <font key="font" metaFont="message"/>
@@ -128,7 +128,7 @@
                     </popUpButtonCell>
                 </popUpButton>
                 <button verticalHuggingPriority="750" translatesAutoresizingMaskIntoConstraints="NO" id="iZs-U3-CJx">
-                    <rect key="frame" x="232" y="96" width="225" height="16"/>
+                    <rect key="frame" x="232" y="72" width="225" height="16"/>
                     <buttonCell key="cell" type="check" title="ESC key clears entire input buffer" bezelStyle="regularSquare" imagePosition="left" state="on" inset="2" id="04m-pr-CAi">
                         <behavior key="behavior" changeContents="YES" doesNotDimImage="YES" lightByContents="YES"/>
                         <font key="font" metaFont="system"/>
@@ -154,7 +154,7 @@
                     </textFieldCell>
                 </textField>
                 <textField focusRingType="none" verticalHuggingPriority="750" translatesAutoresizingMaskIntoConstraints="NO" id="3oc-Wl-pem">
-                    <rect key="frame" x="93" y="224" width="128" height="16"/>
+                    <rect key="frame" x="93" y="192" width="128" height="16"/>
                     <textFieldCell key="cell" scrollable="YES" lineBreakMode="clipping" sendsActionOnEndEditing="YES" alignment="right" title="Candidate Text Size:" id="X0q-Fl-zzQ">
                         <font key="font" metaFont="system"/>
                         <color key="textColor" name="controlTextColor" catalog="System" colorSpace="catalog"/>
@@ -186,8 +186,16 @@
                         <binding destination="32" name="selectedTag" keyPath="values.UseHorizontalCandidateList" id="o5T-Km-VhT"/>
                     </connections>
                 </matrix>
+                <textField focusRingType="none" verticalHuggingPriority="750" translatesAutoresizingMaskIntoConstraints="NO" id="HzV-qQ-eeF">
+                    <rect key="frame" x="67" y="228" width="154" height="16"/>
+                    <textFieldCell key="cell" scrollable="YES" lineBreakMode="clipping" sendsActionOnEndEditing="YES" alignment="right" title="Composition Preference:" id="zFa-xQ-S5v">
+                        <font key="font" metaFont="system"/>
+                        <color key="textColor" name="controlTextColor" catalog="System" colorSpace="catalog"/>
+                        <color key="backgroundColor" name="controlColor" catalog="System" colorSpace="catalog"/>
+                    </textFieldCell>
+                </textField>
                 <popUpButton verticalHuggingPriority="750" translatesAutoresizingMaskIntoConstraints="NO" id="AvT-mU-HNg">
-                    <rect key="frame" x="232" y="217" width="65" height="24"/>
+                    <rect key="frame" x="232" y="184" width="64" height="24"/>
                     <popUpButtonCell key="cell" type="push" title="18" bezelStyle="rounded" alignment="left" lineBreakMode="truncatingTail" state="on" borderStyle="borderAndBezel" tag="18" imageScaling="proportionallyDown" inset="2" selectedItem="64X-UF-Vz6" id="Bsy-hs-5q4">
                         <behavior key="behavior" lightByBackground="YES" lightByGray="YES"/>
                         <font key="font" metaFont="message"/>
@@ -212,7 +220,7 @@
                     <rect key="frame" x="20" y="475" width="438" height="5"/>
                 </box>
                 <box horizontalHuggingPriority="234" verticalHuggingPriority="750" boxType="separator" translatesAutoresizingMaskIntoConstraints="NO" id="Frc-9z-VZe">
-                    <rect key="frame" x="20" y="206" width="445" height="5"/>
+                    <rect key="frame" x="20" y="217" width="445" height="5"/>
                 </box>
                 <button verticalHuggingPriority="750" translatesAutoresizingMaskIntoConstraints="NO" id="9MN-II-LQm">
                     <rect key="frame" x="232" y="339" width="188" height="16"/>
@@ -225,7 +233,7 @@
                     </connections>
                 </button>
                 <textField focusRingType="none" verticalHuggingPriority="750" translatesAutoresizingMaskIntoConstraints="NO" id="Sp0-2u-7hi">
-                    <rect key="frame" x="101" y="178" width="120" height="16"/>
+                    <rect key="frame" x="101" y="154" width="120" height="16"/>
                     <textFieldCell key="cell" scrollable="YES" lineBreakMode="clipping" sendsActionOnEndEditing="YES" alignment="right" title="Shift + Letter Keys:" id="ZFO-C0-adO">
                         <font key="font" metaFont="system"/>
                         <color key="textColor" name="controlTextColor" catalog="System" colorSpace="catalog"/>
@@ -233,7 +241,7 @@
                     </textFieldCell>
                 </textField>
                 <matrix verticalHuggingPriority="750" tag="1" allowsEmptySelection="NO" translatesAutoresizingMaskIntoConstraints="NO" id="s4T-Mw-U2e">
-                    <rect key="frame" x="232" y="156" width="227" height="38"/>
+                    <rect key="frame" x="232" y="132" width="227" height="38"/>
                     <color key="backgroundColor" name="controlColor" catalog="System" colorSpace="catalog"/>
                     <size key="cellSize" width="227" height="18"/>
                     <size key="intercellSpacing" width="4" height="2"/>
@@ -258,7 +266,7 @@
                     </connections>
                 </matrix>
                 <textField focusRingType="none" verticalHuggingPriority="750" translatesAutoresizingMaskIntoConstraints="NO" id="YIg-XJ-f3T">
-                    <rect key="frame" x="162" y="96" width="59" height="16"/>
+                    <rect key="frame" x="162" y="72" width="59" height="16"/>
                     <textFieldCell key="cell" scrollable="YES" lineBreakMode="clipping" sendsActionOnEndEditing="YES" alignment="right" title="ESC Key:" id="Apl-bu-Zdz">
                         <font key="font" metaFont="system"/>
                         <color key="textColor" name="controlTextColor" catalog="System" colorSpace="catalog"/>
@@ -291,7 +299,7 @@
                     </connections>
                 </matrix>
                 <button translatesAutoresizingMaskIntoConstraints="NO" id="cRR-T6-aMw">
-                    <rect key="frame" x="232" y="57" width="155" height="16"/>
+                    <rect key="frame" x="232" y="50" width="155" height="16"/>
                     <buttonCell key="cell" type="check" title="Beep upon input error" bezelStyle="regularSquare" imagePosition="left" alignment="left" state="on" inset="2" id="rnd-fr-0dm">
                         <behavior key="behavior" changeContents="YES" doesNotDimImage="YES" lightByContents="YES"/>
                         <font key="font" metaFont="system"/>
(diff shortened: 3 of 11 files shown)


ChiahongHong:
Although I’m not a reviewer, I’d like to share a thought. If we want to add this behavior, I think it should be handled at the beginning of `walk()`: return the full-input match directly when there is no override, and otherwise use the existing Viterbi flow. Since candidate selection creates an overridden node, we can check the grid directly instead of inspecting `result.nodes` afterward. Running Viterbi first and then replacing its result means that computation is discarded in the exact-match case, so this part may be worth another look.

xatier:
Thanks for implementing the feature, `好險 -> 好顯` is also another common one that bothered me frequently.

lukhnos:
@skyeyvapor 謝謝你發送這份 PR，也謝謝你用心增加 UI 設定！

目前 PR 的作法，開啟設定後，在組字區為空，打出問題詞時，會照 PR 的作法，選擇長字詞。但如果繼續打字打下去，就會回到原有的組字行為，例如「好險沒事，也好險沒有後遺症」，在打到第二個好險時，兩個詞都會變回「好顯」。這樣的使用體驗，似乎對於習慣打長句的用戶，會造成錯愕。不知道目前 PR 這樣的設計，有什麼用戶體驗上的考量嗎？

另外 @ChiahongHong 的留言也值得參考：

> Although I’m not a reviewer, I’d like to share a thought. If we want to add this behavior, I think it should be handled at the beginning of walk(): return the full-input match directly when there is no override, and otherwise use the existing Viterbi flow. Since candidate selection creates an overridden node, we can check the grid directly instead of inspecting result.nodes afterward. Running Viterbi first and then replacing its result means that computation is discarded in the exact-match case, so this part may be worth another look.

我這邊的觀察跟 code review 如下：

「雙字詞因為分數不及同音的兩個單字，而無法變成預設選詞」的問題，例如「好險」（如 @xatier 提及，預設出現「好顯」）跟「面試」（預設為「面是」），確實是個受小麥 unigram model 以及語料來源等因素所造成的困擾。我們其實曾經利用這個 script 分析過：

```
# 更改 109 行的 insufficients_map[2][:25] 到 :100 就會顯示前 100 名的此類詞彙
McBopomofo/Source/Data$ python3 scripts/analyze_data.py
```

我個人的想法是，如果我們採用這個 PR，我們應該將這項變更視為過渡性的作法。至於要從源頭解決這問題，得考慮從資料面或是資料庫製作過程，加入一個「後製」的 boosting stage。我目前的考量—─這也是我們一直以來在這個問題上態度保守的原因──是要判別什麼樣的詞是可以安全的加分、而不會造成 (1) 蓋過真正常用的兩個同音單字詞，或是 (2) 蓋過其他同音二字或三字詞，或是（這也是最大的挑戰）如何避免加分後，原本涵蓋高分單字詞的前後詞組（舉個例子：「確實是也沒錯啦」會不會變成「確實視野沒錯啦」？），結果反而變成輸入體驗上開倒車。

另一個我提議當過渡作法的原因是，我希望這是個使用者不需要思考或發現的功能──換句話說這應該是個從根源著手的問題。這比較屬於設計原則。如果長詞本該為預設詞，理想狀況當然是直接幫使用者做好這樣的設想。另一方面，這個 PR 應該立即可以改進許多常見的詞彙輸入問題，是個 local optimum，但我也好奇想知道有什麼樣的風險或是體驗上 regression 的可能性。

skyeyvapor:
以「好險沒事，也好險沒有後遺症」為例，目前 PR 的確只改善「整個組字區剛好是完整詞」的情況；一旦繼續輸入到「好險沒」，因為 full-input exact match 不再成立，行為就會回到既有 Viterbi/unigram scoring。因此這個 PR 並不會把「好險」在長句中視為全域 boost；在沒有使用者覆寫記憶介入的情況下，長句結果仍會接近目前既有行為，也就是「好顯沒是，也好顯沒有後遺症」。

這是刻意把影響範圍限制在最小的過渡作法：它可以改善像「好險」、「面試」這類單獨輸入時明顯不自然的結果，但避免在長句中任意 boost 二字詞，造成新的斷詞 regression。也就是說，它不是從根源解決 unigram model 的限制，而是比較保守的 local improvement。

我同意這不是最理想的 UX。若要讓長句也自然得到「好險沒事，也好險沒有後遺症」，比較根本的方向可能需要 context-aware scoring，例如 n-gram model 或 reranking，而不只是目前 unigram lattice scoring。這樣模型才有機會根據前後文判斷「好險沒事」比「好顯沒是」自然，同時也避免單純把某些二字詞全域加分後，在其他 context 造成新的斷詞 regression。

我也同意資料面或資料庫製作流程中的後製 boosting 會是比較可落地的中間方向，所以我把目前 PR 定位成比較保守的過渡作法：只在整個 composing buffer 剛好是一個已知詞時偏好該詞，不把同一個詞在長句中全域 boost。長句中的改善仍主要仰賴既有 UserOverrideModel；若要從根源改善，應該另開較完整的資料/modeling 方向來評估。

我也發現之前改動造成了一個跟 UserOverrideModel 互動相關的 regression，因此另外補了一個修正：當 exact-match preference 開啟時，UOM 會先用 exact-match path 查詢，再 fallback 到正常 Viterbi path。這樣可以保留「字彙」改成「自會」這類 same-reading phrase replacement，也保留「好顯」改成「好險」這類 split-word correction，避免使用者覆寫記憶被 exact-match shortcut 蓋掉。

zonble:
如果目標是優先使用二字詞，而不是兩個分數很高的字，我覺得另一種比較直觀的作法是去改 McBopomofoLM，動態將兩字或更長的詞的分數調高。目前這個功能改了 grid 以及 key handler，我覺得修改點有點分散，而且小麥裡頭用了兩個 grid，一個放在 key handler 中，一個放在 service 中，這邊的改動好像沒考慮到service 那還有一份 grid。

目前在 McBopomofoLM 的 filterAndTransformUnigrams 這段，我們會做 unigram 的 value 的轉換，像是替代 macro 的內容、簡繁轉換等，這邊應該也可以放一個 C++ anonymous function 調整分數，這樣改動的地方會比較集中，不用動到 walk 的演算法，而且還可以對內建詞庫還是使用者詞庫分開調整。

skyeyvapor:
我看了一下目前的呼叫鏈，McBopomofoLM 這層目前無法知道某個 key 是否代表整個 composing buffer。ReadingGrid::update() 會對 lattice 裡所有可能 span 呼叫 getUnigrams(combinedReading)，所以對 McBopomofoLM 來說，getUnigrams("zi-hui") 無法分辨它是完整輸入，還是長句中的 prefix / middle span。

因此如果要把這個行為放到 McBopomofoLM，可能需要擴充 LanguageModel API，把 span context 或 isFullInput 傳下去；否則在 filterAndTransformUnigrams 裡調分會變成較全域的 boosting，語意就不再是目前 PR 的 full-input-only local improvement。

所以若目標是維持 local improvement，我目前仍傾向把 full-input 判斷留在 ReadingGrid，或改成 walk options 讓呼叫端更明確；如果要改成 LM scoring boost，那可能會是另一個較大 scope 的 scoring policy 討論。

tianjianjiang:
## TL;DR — Proposed path

Every example in this PR was verified against `data.txt` and the MOE dictionary (verification scripted by Claude Code). They are not one problem but four:

- (A) a lexeme losing to a homophonous non-lexical character string (好險 vs 好+顯, 面試 vs 面+是, ...);
- (B) homophonous entries in genuine competition (不已 vs 不以, 步道 vs 不到, 試試 vs 逝世/事事/世事);
- (C) defects in `data.txt` entries (一在 exists as an entry; 得出 is filed under the particle reading ㄉㄜ˙ while 得到 is under ㄉㄜˊ);
- (D) inconsistent coverage of tone-sandhi and neutral-tone syllable strings, which silently splits users into protected and unprotected groups (不到 exists only under ㄅㄨˊ; 鑰匙 under ㄕ˙, ㄕˊ and ㄕˇ).

One more observation on (A): a handful of characters dominate it — 是 alone is involved in six of the seventeen examples — so whatever we learn about those few weights will pay off across the whole category. Each category needs a different treatment, so instead of fixing one word at a time:

1. Write a Python replica of the engine's `walk()` over `data.txt`.
2. Verify that the replica and the engine always select identically.
3. Add the `promote`/`assert` test format lukhnos sketched earlier.
4. Seed a corpus from this PR's examples plus the guard sentences any score change must not break.
5. Gate `data.txt` changes in CI with that corpus.
6. Land the clear-cut category-(A) fixes first — one-line data changes whose safety margins the corpus has verified.
7. For the rest of (A) and all of (B), evaluate candidate algorithms on the same corpus and let one report decide (candidates in a later section).
8. Investigate why a few single-character weights dominate (A), against the corpus statistics `data.txt` was built from, before deciding whether re-estimation is warranted.
9. Audit (C) and (D) across all of `data.txt`, each fix gated by the corpus.

This PR stands as a transitional fix for the whole-buffer exact-match case; none of the above conflicts with it.

## The cases, category by category

### (A) Lexemes losing to non-lexical character strings

The cases: 好險→好顯, 面試→面是, 鑰匙→要時, 依舊→一就, 做事→作是, 視野→是也, 各式→個是, 好事→好是, 心事→心是, 事後→是後, 有事→有是, 試試→是是, plus 沒事→沒是, which is not in the PR table — it was found while verifying 好險沒事, so systematic scanning beats collecting reports. Margins (the lexeme's log probability minus the winning split's) range from −0.004 (好險, the narrowest) to −0.834 (有事).

The substantial facts:

- 鑰匙 loses only under ㄧㄠˋ-ㄕˊ; the PR table cites ㄧㄠˋ-ㄕ˙, where 鑰匙 already wins — see (D).
- Positional distribution matters throughout: a classifier such as 個 cannot start an utterance, 不已 occurs only after a verb — and the unigram model cannot condition on position. This dimension recurs in (B).
- Several winning strings have legitimate chunk-boundary sources — 這/是不是/也/一樣, 三個/是紅的, 約好/是七點 — so the guard corpus must contain such sentences before any weight moves.
- 做事, 心事, 事後 and 試試 share their syllable strings with other lexemes (坐視/作勢; 新式/心室/新市; 侍候; 逝世/事事/世事) — a (B) overlap: those changes must be judged by the whole candidate list.
- 字彙→自會, 一再→一在 and 得出→的出 present as (A) but each sits on a `data.txt` entry defect — see (C).

### (B) Homophonous entries in genuine competition

Here the current output is not a mistake in any absolute sense — every contender is a deliberate `data.txt` entry — so the question is not "which entry is wrong" but "what information could tell the intents apart".

- 不已 vs 不以: both MOE lexemes (不已 「不絕、不停止」; 不以 「不因、不因為」). Their positional distributions are nearly complementary: 不已 occurs after a verb (感激不已), 不以 opens a clause (不以成敗論英雄). Today's output 不以 actually comes from the character split 不+以, which happens to spell the same string. The distinguishing information is position, which the unigram model cannot see.
- 步道 vs 不到: both MOE lexemes (MOE files 不到 under the citation tones ㄅㄨˋ ㄉㄠˋ). `data.txt` carries 不到 only under the sandhi syllable string ㄅㄨˊ-ㄉㄠˋ, so users who type the sandhi tone put the two intents on different syllable strings and never collide, while users who type citation tones put both intents on ㄅㄨˋ-ㄉㄠˋ — for them even the standalone input is genuinely ambiguous (「一百嗎?」「不到。」 is a complete answer). This case interlocks with the syllable-string coverage issues in (D).
- 他的 vs 它的 vs 她的: none of the three is an MOE lexeme — they are compositional pronoun+的 strings carried as deliberate convenience entries, ranked 他的 > 它的 > 她的 today. The distinguishing information is the antecedent's gender, which no amount of score tuning can encode. What matters operationally is that the existing ranking is meaningful, and any future change on any syllable string should preserve same-string order unless it has evidence to reorder it — the concern lukhnos raised on Slack.
- The family syllable strings from (A) — 試試/逝世/事事/世事, 心事/新式/心室/新市, 做事/坐視/作勢, 事後/侍候: the reported symptom is the (A) split winning, but these syllable strings each host several entries, so any change must be judged by the whole candidate list, not only by which entry ranks first.

### (C) Defects in `data.txt` entries

These present as (A)-style losses, but each traces back to an entry whose existence or syllable string is in question — the entry must be judged before any score is.

- 一在: an entry under both the ㄧ- and ㄧˊ- syllable strings, with no MOE lexeme behind it; 一再 (an MOE lexeme) loses to it. While it stands, tuning 一再 aims at a moving target.
- 自會: an entry (there is even a 自會在) with no MOE lexeme — why 字彙 was referred here from (A). If 自會 is judged compositional, 字彙 reduces to an ordinary (A) case against 自+會.
- 得出: exists only under ㄉㄜ˙-ㄔㄨ, losing to 的+出; nothing under ㄉㄜˊ-ㄔㄨ, though 得到 sits under ㄉㄜˊ-ㄉㄠˋ. MOE has no 得出 lexeme, but it does fix the character: verb 得 is ㄉㄜˊ, ˙ㄉㄜ is the particle. The sense users lose (得出結論) is the verb, so the syllable string it needs has no entry; whether the ㄉㄜ˙ entry is that verb misfiled or a V-得-出 fragment (cf. 提得出 under both syllable strings), the defect is in the entry inventory, not the weights.
- 匙: a standalone single-character entry under ㄕ˙, yet in that reading 匙 is a bound morpheme — MOE defines ˙ㄕ solely by cross-reference to 鑰匙, and it never occurs as a free form; the free morpheme is 匙 ㄔˊ. A bound morpheme cannot stand alone as a candidate, so the entry is a phantom on every walk crossing ㄕ˙ — though it does not decide the 鑰匙 case itself (鑰匙 wins there).

All four surfaced incidentally while verifying this PR's examples — samples, not the population. A whole-file audit of `data.txt` should flag entries MOE does not attest and bound morphemes promoted to standalone entries, as review flags rather than verdicts, since `data.txt` legitimately carries words beyond MOE.

### (D) Inconsistent coverage of tone-sandhi and neutral-tone syllable strings

Unlike (C), every entry here is a genuine lexeme; the defect is which syllable strings carry it. Users differ in whether they type tone sandhi and neutral tones, and `data.txt`'s coverage of those variants follows no discernible policy — so which habit a user has silently decides whether an entry protects them. The examples span the full range:

- 不到: only under the sandhi form ㄅㄨˊ-ㄉㄠˋ, while MOE cites ㄅㄨˋ ㄉㄠˋ. Sandhi typists get 不到 on a syllable string of its own; citation-tone typists land on ㄅㄨˋ-ㄉㄠˋ, where only 步道 exists and the 不到 intent must win as the split 不+到 — which it does, at 步道's expense (the (B) case). The 步道/不到 collision exists for one user group and not the other.
- 心事 and 試試: MOE's readings are ㄒㄧㄣ ˙ㄕ and ㄕˋ ˙ㄕ — neutral second syllable — yet both are carried only under full ㄕˋ. Users typing MOE's own reading reach no entry at all; the (A) losses are the *better* of their two situations.
- 本事: both MOE readings (ㄅㄣˇ ˙ㄕ, ㄅㄣˇ ㄕˋ) have syllable strings at the same weight — the coverage 心事 and 試試 lack.
- 鑰匙: three syllable strings — ㄧㄠˋ-ㄕ˙ (MOE's reading, where it wins) plus full-tone variants ㄧㄠˋ-ㄕˊ (where it loses — the real site of the PR's reported case; the table's ㄕ˙ citation is addressed in (A)) and ㄧㄠˋ-ㄕˇ, all at the same weight. The most generous coverage in the set, yet the protection does not extend across the variant family.

One entry carried only under sandhi, two carried only against MOE's neutral tone, one carried under both, one carried under three syllable strings with uneven outcomes: the defect is not any single syllable string but the absence of a policy for which tone variants an entry must carry — and until that is explicit, every per-word fix risks repairing one user group while leaving the other where it was.

## Candidate algorithms for the rest of (A) and all of (B)

This is TL;DR step 7 in detail. After the clear-cut (A) nudges land, the remaining cases need information the unigram model does not carry. Five candidates will run on the same replica walk and the same corpus, and the comparison report decides; nothing below is a commitment. Four change the scoring; the fifth changes the weights themselves — the harness judges outputs either way.

- Length prior (λ), i.e. a word insertion penalty (Donaj & Kačič): language models structurally favor splits into fewer, shorter units; speech recognition corrects this with a fitted per-word bonus — a significant 2.02% relative word error rate reduction, but only after corpus optimization. Here it becomes a uniform per-syllable bonus on entries; uniform addition cannot reorder entries sharing a syllable string, so the (B) order invariant (他的 > 它的 > 她的) holds by construction. The sweep already ran: λ ≤ 0.20 fixes 好險, 鑰匙 (ㄧㄠˋ-ㄕˊ), 面試 and 好事 with the guard set green; λ = 0.22 also fixes 步道 but breaks 等不到人; λ = 0.28 fixes 視野 but breaks 這是不是也一樣. So λ alone cannot cover the set, and with only 17 guard sentences the "safe" window is a floor on risk — fitted, not hand-picked, as in the source literature.
- Three-word chunking (Chen & Liu 1992): on ambiguity, form every chunk of the next three words and commit the best chunk's first word. Best = greatest total length: in wan-cheng-jian-ding-bau-gau, 完成/鑑定/報告 is the longest chunk, so 完成 wins over 完 — this one rule resolved 93% of their ambiguities at 99.7% accuracy. Ties fall to the most even word lengths (研究生命起源 → 研究/生命/起源, not 研究生/命/起源), then to everyday monosyllabic function words over rare ones. Their own caution: scoring a chunk by adding word frequencies was their weakest rule, under 70% — pure frequency accumulation without structural preference fails on exactly these local ambiguities.
- Dual-rank resolution (Jiang, Lee & Hsu 2013): one frequency cannot express a word's strength in both directions. 軍事 wins the overlap on its right (jun-shi-jie → 軍事/界) but loses on its left (lu-jun-shi → 陸軍/是), and the paper's examples include 家事 vs 是 — our 事/是 family. So each word carries two fitted ranks; at an overlap, compare the left word's right-rank with the right word's left-rank, tie → frequency. That is exactly the positional asymmetry in our cases: 不已 after verbs, 不以 clause-initial. The fitting matters as much as the rule: ranks come from a competition graph (words sharing a boundary syllable; arcs weighted by corpus preference) solved as minimum feedback arc set, and the arcs no assignment can satisfy are internal conflicts of the data itself. Built over `data.txt` and the guard corpus, that check would tell the regression tool in advance which assertions cannot all hold — λ = 0.22 fixing 步道 while breaking 等不到人 is such an arc pair, found today only by trial.
- Term-contributed frequency re-estimation (Sung et al. 2008, via their YASA suffix array): every 但是, 就是 or 是否 in the corpus also counts as one 是, so 是's standalone weight earns credit for strings no user types 是 alone in. TCF counts only what a string contributes by itself, discounting occurrences absorbed by a longer string. 是 sits in seven (A) cases (counting 沒事); if the inflation is real, all their margins shrink at once with no per-word tuning. Precondition: the statistics of the source corpus the weights came from (TL;DR step 8) — the inflation is still a hypothesis, not a demonstrated cause.
- Hybrid: λ for the length axis, a contextual candidate for the position axis, and TCF composing with either as a weights change — the axes are largely orthogonal in the cases above.

Provenance: the margins and λ figures above come from a local, uncommitted Python replica of the engine's `walk()` over `data.txt`, checked by a small assertion corpus seeded from this PR's examples (28 checks pass on unmodified data; 9 sentences await native-speaker review). The replica is not yet cross-checked against gramambular2, so these are preliminary measurements, not controlled experiments. The tooling, including lukhnos's test format, will be proposed in its own issue and PR.

## Proposed follow-ups

Each TL;DR step becomes its own issue — the regression tool (steps 1–5), the corpus-gated nudges (step 6, 好險 first as the narrowest margin), the five-candidate comparison (step 7, after the cross-check), the single-character weight investigation (step 8), and the (C)/(D) audits (step 9). Three items the TL;DR does not already cover:

1. Native-speaker review of the 9 sentences from the provenance note: they were generated by Claude Code (e.g. 等不到人), so before they can veto a scoring change, a native speaker must confirm they are natural.
2. The (C) audit's flags: entries MOE does not attest, and bound morphemes carried standalone.
3. The (D) audit's prerequisite: a tone-variant coverage policy.

None of this blocks the present PR.

## References

- Chen, K.-J., & Liu, S.-H. (1992). Word Identification for Mandarin Chinese Sentences. COLING-92. <https://aclanthology.org/C92-1019.pdf>
- Donaj, G., & Kačič, Z. (2013). The Use of Several Language Models and Its Impact on Word Insertion Penalty in LVCSR. <https://doi.org/10.1007/978-3-319-01931-4_47>
- Jiang, M. T.-J., Lee, T.-H., & Hsu, W.-L. (2013). The Left and Right Context of a Word: Overlapping Chinese Syllable Word Segmentation with Minimal Context. ACM TALIP. <https://doi.org/10.1145/2425327.2425329> (open access: <https://iasl.iis.sinica.edu.tw/webpdf/paper-2012-The_Left_and_Right_Context_of_a_Word_Overlapping_Chinese_Syllable_Word_Segmentation_with_Minimal_Context.pdf>)
- Sung, C.-L., Yen, H.-C., & Hsu, W.-L. (2008). Compute the Term Contributed Frequency. (as cited in Jiang et al., 2012, IJCLCLP, 17(3), 45–86, <https://aclanthology.org/O12-4003.pdf>)
- MOE dictionary entries were checked via moedict: <https://www.moedict.tw/>

---

*Posted by Claude Code on behalf of @tianjianjiang.*

tianjianjiang:
上面那篇是我跟 Claude Code 整理了幾小時的草案。裡面還是有一些如果我手工寫的話不會寫成那樣的奇怪說法，但大方向上應該還過得去。

我想，接下來就看各方面的進度來決定我可以幫忙做什麼。

tianjianjiang:
附帶一提，

> 無法分辨它是完整輸入，還是長句中的 prefix / middle span。

我印象中應該有 `<s>` 這個特殊的 unigram 標記是不是開頭。實際上現在的 dynamic programming 怎麼管理這類情況我還沒檢查。

另外，要是這個 PR 的目標是讓二字詞在 local 裡盡可能地正確，那麼我推薦只先做真的是詞的部分。

ChiahongHong:
Hi @tianjianjiang 不好意思好奇想問，您提到的 `<s>` 是指 SRILM 的格式嗎～ 我記得會有 `<s>` 與 `</s>` 代表 BOS 與 EOS (bigram 的話)，不過小麥的 unigram 應該沒有這樣子的標記，目前 Viterbi DP 也沒有相關的處理機制。

zonble:
> 我看了一下目前的呼叫鏈，McBopomofoLM 這層目前無法知道某個 key 是否代表整個 composing buffer。ReadingGrid::update() 會對 lattice 裡所有可能 span 呼叫 getUnigrams(combinedReading)，所以對 McBopomofoLM 來說，getUnigrams("zi-hui") 無法分辨它是完整輸入，還是長句中的 prefix / middle span。

您好：

其實我不是很懂您想要的效果，或是說，我不是很懂「一定要代表整個 composing buffer」的理由。

我聽起來是，您希望剛好打下「ㄗˋ-ㄏㄨㄟˋ」兩個字之後，出現「字彙」，用戶就要按下 enter 鍵送出嗎？這樣的確就像 lukhnos 前面講的，如果繼續往下打，那又會從「字彙」變成「字會」了。

zonble:
我在 Web 版那邊先簡單做了點小實驗。

https://github.com/openvanilla/McBopomofoWeb/commit/edb6a6d0f249692a59a06b0584bf5fb43d2bca81

大慨就是簡單的增加權重，然後愈長的詞權重就會動態增加到愈高。

skyeyvapor:
是的，我這個 PR 的目標比較保守：希望使用者剛好打下「ㄗˋ-ㄏㄨㄟˋ」這類完整讀音時，如果整個 composing buffer 本身就是一個已知詞，就先顯示「字彙」，讓使用者可以直接按 Enter 送出。

這比較像是一個 local workaround，主要是針對我自己也常遇到、而且滿干擾的情況：單獨輸入一個常見詞時，結果卻被切成兩個高分單字。它不是想從根本上改變長句斷詞，也不是想讓這個詞在所有 context 裡都被 boost。

所以我刻意把條件限制在 full composing buffer exact match。這樣的缺點確實是，如果使用者繼續往下打，結果可能又會回到目前既有的 Viterbi/unigram scoring，例如前面提到的長句情境。但好處是影響範圍比較小，比較不容易改壞像「等不到人」、「這是不是也一樣」這類原本合理的句子。

我也同意 Web 版那種依詞長動態加權的方向比較接近真正改善斷詞品質。不過那會影響所有 lattice 裡的候選詞，包含長句中的局部片段，所以我會比較擔心 regression。若要走那個方向，我覺得最好先搭配 corpus / guard sentence 測試，再決定加權幅度。

PR #869, opened 2026-07-06
Fix the emoji only readings

zonble:
The PR updates `analyze_data.py` to find the readings that only map to emojis.

Also updated the phrase database to fix such readings.

The PR ever added scripts to find the phrases whose score is lower than the sum of the components. The script is removed since it is duplicated with `analyze_data.py`.


Source/Data/scripts/analyze_data.py
```diff
@@ -6,14 +6,21 @@
...
 unigram_1char = {}
+unigram_emoji = {}
 value_to_score = {}
 
 unigram_1char_count = 0
 unigram_multichar_count = 0
+emoji_count = 0
 
 for (r, v, s) in data:
     # Skip emojis.
```
tianjianjiang, line 18: `if s == -8` selects emojis by their score sentinel, but 49 `MACRO@…` date/time entries share the same −8 score, so they are counted as emojis and pushed into `unigram_emoji`. On the built dictionary this makes the summary report 1496 emojis when there are 1447, and the macros then flow into the final "not typed by any unigram or phrase" check (they avoid surfacing today only because their readings happen to match real phrase readings). Gating on the value shape fixes both:

```suggestion
    # Skip macros; they share the -8 score with emojis but are not emojis.
    if v.startswith('MACRO@'):
        continue
    # Skip emojis.
```

*Posted by Claude Code on behalf of @tianjianjiang.*

Source/Data/BPMFMappings.txt
```diff
@@ -9240,6 +9244,7 @@
 亞哥花園 ㄧㄚˇ ㄍㄜ ㄏㄨㄚ ㄩㄢˊ
 亞喬木 ㄧㄚˇ ㄑㄧㄠˊ ㄇㄨˋ
 亞塞拜然 ㄧㄚˇ ㄙㄜˋ ㄅㄞˋ ㄖㄢˊ
+亞塞拜然 ㄧㄚˋ ㄙㄞˋ ㄅㄞˋ ㄖㄢˊ
```
tianjianjiang, line 9249: This PR adds `亞塞拜然 ㄧㄚˋ ㄙㄞˋ` — completing the 亞×塞 reading matrix lets every attested reading resolve to the phrase.

亞 and 塞 are both polyphonic here, and the reading sources split between prescriptive and usage:
- **亞**: ㄧㄚˋ (MOE 重編辭典 [ID 152278](https://dict.revised.moe.edu.tw/dictView.jsp?ID=152278&q=1&word=亞塞拜然)) / ㄧㄚˇ ([moedict 亞](https://www.moedict.tw/uni/亞) 又音; [Wiktionary](https://en.wiktionary.org/wiki/亞塞拜然語) variant)
- **塞**: ㄙㄜˋ (MOE ID 152278) / ㄙㄞˋ ([moedict 塞](https://www.moedict.tw/uni/塞) 要塞 reading; Wiktionary gives *only* ㄙㄞˋ)

Four attested combinations. The file has ㄧㄚˇ-ㄙㄜˋ (pre-existing) and ㄧㄚˋ-ㄙㄞˋ (this PR, = Wiktionary's *standard*). Adding ㄧㄚˋ-ㄙㄜˋ (MOE reading — otherwise a user typing the dictionary reading can't produce the phrase) and ㄧㄚˇ-ㄙㄞˋ (Wiktionary variant) covers the rest. Additive to `BPMFMappings.txt` only — both map to the existing `亞塞拜然` phrase, no `phrase.occ`/`PhraseFreq.txt` change. The block below is already in `LC_ALL=C` order, so `make sort` stays a no-op:

```suggestion
亞塞拜然 ㄧㄚˇ ㄙㄜˋ ㄅㄞˋ ㄖㄢˊ
亞塞拜然 ㄧㄚˇ ㄙㄞˋ ㄅㄞˋ ㄖㄢˊ
亞塞拜然 ㄧㄚˋ ㄙㄜˋ ㄅㄞˋ ㄖㄢˊ
亞塞拜然 ㄧㄚˋ ㄙㄞˋ ㄅㄞˋ ㄖㄢˊ
```

*Posted by Claude Code on behalf of @tianjianjiang.*

Source/Data/scripts/analyze_data.py
zonble: Fixed in 2d71b4cea3b43974aace440120ba3e7de8eaafd1

Source/Data/BPMFMappings.txt
zonble: Fixed in 8ffd48ad834becfec2ab78b82e142ea55d557208

lukhnos (approved):
LGTM. Thanks!

PR #883, opened 2026-07-18
Add engine profiler

ChiahongHong:
Since all of our current performance benchmarks are based on unit tests, it has been difficult to identify bottlenecks across the end-to-end input flow. This PR adds an engine profiling workload that simulates real-world input and uses `xctrace` to generate `.trace` files that can be opened directly in Instruments.

```
Source/Tools/run-engine-profiler.sh
```

The report will be generated in `Source/Engine/Report/`.

Without NEON:
```
short_iterations=771129
medium_iterations=116411
long_iterations=30573
```

With NEON:
```
short_iterations=898449
medium_iterations=148747
long_iterations=43103
```

PS: short, medium, and long refer to the length of the bopomofo sequence in each `ProfilingScenario`.

The traces confirm that binary search is indeed the largest performance cost. After the recent PRs introduced SIMD optimizations, I wanted to approach the problem from another angle: can we reduce the number of binary searches performed in the first place?

For example, when entering `ㄒㄧㄠˇ ㄇㄞˋ ㄓㄨˋ ㄧㄣ ㄕㄨ ㄖㄨˋ ㄈㄚˇ`, the lookup results show a significant number of misses. As each new reading is entered, many of the same missing keys are searched again:

```
insert[1] keys=vul3 reading=ㄒㄧㄠˇ
  query[1] reading=ㄒㄧㄠˇ unigram_count=9
    value=小 (...remaining values omitted)
insert[2] keys=a94 reading=ㄇㄞˋ
  query[2] reading=ㄒㄧㄠˇ-ㄇㄞˋ unigram_count=1
    value=小麥
  query[3] reading=ㄇㄞˋ unigram_count=14
    value=賣 (...remaining values omitted)
insert[3] keys=5j4 reading=ㄓㄨˋ
  query[4] reading=ㄒㄧㄠˇ-ㄇㄞˋ-ㄓㄨˋ unigram_count=0
  query[5] reading=ㄇㄞˋ-ㄓㄨˋ unigram_count=0
  query[6] reading=ㄓㄨˋ unigram_count=60
    value=助 (...remaining values omitted)
insert[4] keys=up reading=ㄧㄣ
  query[7] reading=ㄒㄧㄠˇ-ㄇㄞˋ-ㄓㄨˋ unigram_count=0
  query[8] reading=ㄒㄧㄠˇ-ㄇㄞˋ-ㄓㄨˋ-ㄧㄣ unigram_count=0
[...]

PR #886, opened 2026-07-19
Avoid repeated negative unigram lookups

ChiahongHong:
Description copied from [#883](https://github.com/openvanilla/McBopomofo/pull/883#issuecomment-5013779351)

---

For caching, I tried keeping an array in each span to record missing keys. I also tried using an `unordered_map` to remember which keys had already been looked up. I know 新酷音 uses a trie, but that feels too heavy for this. None of these approaches felt elegant enough to me.

In the end, this is the version I am happy with and wanted to share. It does not add any new data structure. The core change is just the loops in `ReadingGrid::update()`.

```cpp
void ReadingGrid::update(size_t loc, EditType editType) {
  // Spans that do not cross the edit retain their previous lookup result. A
  // node means that the lookup succeeded, while a null slot means that the
  // same reading was already looked up and did not exist. Only spans that
  // include an insertion or cross a deletion boundary need to be queried.
  size_t affectedLength = kMaximumSpanLength - 1;
  size_t begin = loc <= affectedLength ? 0 : loc - affectedLength;
  size_t end = editType == EditType::kInsertion ? loc + 1 : loc;
  end = std::min(end, readings_.size());

  for (size_t pos = begin; pos < end; pos++) {
    size_t minimumLength = loc - pos + 1;
    size_t maximumLength = std::min(kMaximumSpanLength, readings_.size() - pos);
    for (size_t len = minimumLength; len <= maximumLength; len++) {
      std::string combinedReading =
          combineReading(readings_.begin() + static_cast<ptrdiff_t>(pos),
                         readings_.begin() + static_cast<ptrdiff_t>(pos + len));

      if (!hasNodeAt(pos, len, combinedReading)) {
        auto unigrams = lm_.getUnigrams(combinedReading);
        if (unigrams.empty()) {
          continue;
        }

        insert(pos, std::make_shared<Node>(std::move(combinedReading), len,
                                           std::move(unigrams)));
      }
    }
  }
}
```

Let's do some simple math XDD, we only want `update()` to visit combinations that can reach the edit location:

```cpp
size_t affectedLength = kMaximumSpanLength - 1;
size_t begin = loc <= affectedLength ? 0 : loc - affectedLength;
```

`kMaximumSpanLength` is 8, so if there are fewer than 7 readings on the left, we start at position 0. Otherwise, we start at `loc - 7`.

The end position depends on whether this is an insertion or deletion:

```cpp
size_t end = editType == EditType::kInsertion ? loc + 1 : loc;
```

For an insertion, `end` is `loc + 1`. The outer loop is:

```cpp
for (size_t pos = begin; pos < end; pos++)
```

Since the condition is `pos < end`, `end` must be `loc + 1` if we want to include `pos == loc`.

Suppose we insert `n` between `a` and `b`:

```text
a b c
```

After the insertion:

```text
a n b c
  ↑
 loc
```

The affected combinations starting before `n` are:

```text
a n
a n b
a n b c
```

The affected combinations starting at `n` are:

```text
n
n b
n b c
```

All of them contain the inserted reading. Existing combinations such as `b c` do not contain the edit, so they can keep their previous lookup result.

For a deletion, `end` is just `loc`.

After deletion, `loc` is the first surviving reading on the right. For example:

```text
a b c
```

After deleting `b`:

```text
a | c
    loc = 1
```

Only spans starting to the left of `loc` can cross this new boundary. So deletion uses `pos < loc` and does not include `pos == loc`.

Then we calculate the minimum length:

```cpp
size_t minimumLength = loc - pos + 1;
```

This is the minimum length needed for a span starting at `pos` to reach `loc`.

The maximum length is:

```cpp
size_t maximumLength =
    std::min(kMaximumSpanLength, readings_.size() - pos);
```

This is the number of readings left between `pos` and the end of the vector, capped by `kMaximumSpanLength`.

That's all!! We've covered all the cases without adding a trie, a cache, or any other long-lived data structure.
My changes are available in https://github.com/ChiahongHong/McBopomofo/tree/reduce-binary-search

### Without NEON

| Stage  | short_iterations | medium_iterations | long_iterations |
|--------|-----------------:|------------------:|----------------:|
| Before |          771,129 |           116,411 |          30,573 |
| After  |          892,077 |           148,430 |          43,351 |

### With NEON

| Stage  | short_iterations | medium_iterations | long_iterations |
|--------|-----------------:|------------------:|----------------:|
| Before |          898,449 |           148,747 |          43,103 |
| After  |          952,787 |           174,237 |          66,322 |

### DIFF
--- a/Source/Engine/gramambular2/reading_grid.cpp
+++ b/Source/Engine/gramambular2/reading_grid.cpp
@@ -54,14 +54,18 @@ bool ReadingGrid::insertReading(const std::string& reading) {
     return false;
   }
 
-  if (!lm_.hasUnigrams(reading)) {
+  std::string insertedReading = reading;
+  auto unigrams = lm_.getUnigrams(insertedReading);
+  if (unigrams.empty()) {
     return false;
   }
 
   readings_.insert(readings_.begin() + static_cast<ptrdiff_t>(cursor_),
-                   reading);
+                   std::move(insertedReading));
   expandGridAt(cursor_);
-  update();
+  insert(cursor_,
+         std::make_shared<Node>(readings_[cursor_], 1, std::move(unigrams)));
+  update(cursor_, EditType::kInsertion);
 
   // Cursor must only move after update().
   ++cursor_;
@@ -78,7 +82,7 @@ bool ReadingGrid::deleteReadingBeforeCursor() {
   // Cursor must decrement for grid-shrinking and update to work.
   --cursor_;
   shrinkGridAt(cursor_);
-  update();
+  update(cursor_, EditType::kDeletion);
   return true;
 }
 
@@ -90,7 +94,7 @@ bool ReadingGrid::deleteReadingAfterCursor() {
   readings_.erase(readings_.begin() + static_cast<ptrdiff_t>(cursor_),
                   readings_.begin() + static_cast<ptrdiff_t>(cursor_ + 1));
   shrinkGridAt(cursor_);
-  update();
+  update(cursor_, EditType::kDeletion);
   return true;
 }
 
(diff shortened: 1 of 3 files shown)


lukhnos (approved):
LGTM.

Thanks again for the contribution!

PR #892, opened 2026-08-06
Keep BPMFBase.txt sorted by reading.

zonble:
The character map was not in a correct order. Fixed.

### DIFF
--- a/Source/Data/BPMFBase.txt
+++ b/Source/Data/BPMFBase.txt
@@ -6715,7 +6715,6 @@
 㱒 ㄏㄜ he ck utf8
 抲 ㄏㄜ he ck utf8
 𠿒 ㄏㄜ he ck utf8
-㗿 ㄏㄜˇ he3 ck3 utf8
 賀 ㄏㄜˋ he4 ck4 big5
 和 ㄏㄜˋ he4 ck4 big5
 鶴 ㄏㄜˋ he4 ck4 big5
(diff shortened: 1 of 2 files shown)


lukhnos (approved):
LGTM. Thanks!

Source/Data/BPMFBase.txt
```diff
@@ -8215,6 +8214,7 @@
 𩿇 ㄍㄨˋ gu4 ej4 utf8
 骨 ㄍㄨˊ gu2 ej6 big5
 鶻 ㄍㄨˊ gu2 ej6 big5
+姑 ㄍㄨ˙ gu5 ej7 big5
```
xatier, line 8217: this should go after `鼓 ㄍㄨ˙`?

xatier:
@lukhnos I'm under the impression that we typically don't adjust the orders in this file? I just noticed that two characters may be put at incorrect order?

Source/Data/BPMFBase.txt
```diff
@@ -11299,6 +11299,7 @@
 藂 ㄘㄨㄥˋ cong4 hj/4 big5
 謥 ㄘㄨㄥˋ cong4 hj/4 big5
 䈡 ㄘㄨㄥˋ cong4 hj/4 utf8
+衳 ㄘㄨㄥˇ cong3 hj/3 utf8
```
xatier, line 11302: this should go after `幒 ㄘㄨㄥˇ`?

Source/Data/BPMFBase.txt
lukhnos: This one turns out to be correct.

Source/Data/BPMFBase.txt
lukhnos: This one also turns out to be correct, but see the other comment of mine.

lukhnos:
> @lukhnos I'm under the impression that we typically don't adjust the orders in this file? I just noticed that two characters may be put at incorrect order?

Good call. After your comement I checked the built `data-plain-bpmf.txt`, and there are three entries whose orders changed. Just made #894 to make sure that the built data for Plain Bopomofo maintains the same order as before.

xatier:
@lukhnos thanks for the prompt reply. I was just wondering why the "new characters" are not "append only". :)

PR #893, opened 2026-08-06
Fix two more incorrect string length calculations

lukhnos:
This is the follow up of #891

cc @ChiahongHong


ChiahongHong:
LGTM, thanks!

zonble (approved):

PR #897, opened 2026-08-22
Support full-width pipe symbol; data clean up

lukhnos:
- Maps the `|` key to the full-width pipe symbol, full-width forward and backward slashes, and half-width pipe symbol at the end so that while it breaks past conventions, it still provides a way for users who need half-width pipe symbols.
- Maps the `|` key in half-width punctuation mode to the half-width pipe symbol and half-width forward and backward slashes. Since there was already an entry for it in the half-width mode, hitting `|` key in half-width mode always causes a symbol on the composing area, and so we might as well add the two other slashes to match the behavior of its full-width counterpart.
- Cleans up all `_kana_` entries since the Iroha mode was superceded by the ICU-powered kana/Hangul conversion mode.


zonble (commented):
LGTM. Thanks!

zonble (approved):

PR #902, opened 2026-08-24
Bump version to 3.1-dev for built-from-source uses

lukhnos:
This is to distinguish users who build from source from those who use the public releases.


mjhsieh (approved):
lgtm

PR #908, opened 2026-08-31
fix(input): refresh candidates after user phrase changes

skyeyvapor:
## Summary

This PR refreshes candidates after a user phrase is added, boosted, or excluded.

## Motivation

After adding `高科` as a user phrase, the language model was reloaded, but the existing `ReadingGrid` nodes still contained the old candidates. As a result, `高科` did not immediately appear in the candidate list.

Similarly, removing `高科` did not immediately remove it from the candidate list.

## Behavior

After the language model is reloaded:

- Nodes matching the changed reading are added, replaced, or removed.
- Every occurrence of the same reading in the grid is refreshed.
- Other nodes, readings, and the cursor remain unchanged.
- The current composition remains unchanged until a later normal walk.
- Valid candidate overrides are preserved.

## Notes

`ReadingGrid::walk()` only selects a path from the nodes already in the grid. It does not query the language model again.

This change refreshes only the nodes matching the updated reading instead of rebuilding the entire grid.

## Testing

Added tests covering:

- Adding and removing `高科` candidates.
- Refreshing repeated readings.
- Preserving the current walk and valid candidate overrides.
- Removing an excluded candidate override.
- Preserving required single-reading nodes.

All 32 `ReadingGridTest` tests pass.

### DIFF
--- a/Source/Engine/gramambular2/reading_grid.cpp
+++ b/Source/Engine/gramambular2/reading_grid.cpp
@@ -98,25 +98,55 @@ bool ReadingGrid::deleteReadingAfterCursor() {
   return true;
 }
 
-bool ReadingGrid::refresh() {
-  std::vector<Span> refreshedSpans(readings_.size());
+bool ReadingGrid::refreshNodesForReading(const std::string& reading) {
+  if (reading.empty()) {
+    return false;
+  }
+
+  struct MatchingSpan {
+    size_t position;
+    size_t length;
+  };
+  std::vector<MatchingSpan> matchingSpans;
   for (size_t pos = 0; pos < readings_.size(); ++pos) {
     size_t maximumLength = std::min(kMaximumSpanLength, readings_.size() - pos);
     for (size_t len = 1; len <= maximumLength; ++len) {
       std::string combinedReading =
           combineReading(readings_.begin() + static_cast<ptrdiff_t>(pos),
                          readings_.begin() + static_cast<ptrdiff_t>(pos + len));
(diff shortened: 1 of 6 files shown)


lukhnos (commented):
Thanks for making this PR, and I appreciate the introduction of `selectedReading` to remove duplicate code.

While I understand what this PR does, could you help us understand what problem or usability issues it tries to address, and what downsides the status quo has?

My assumption has been that an input session is usually short. After a user phrase change, the user moves on to type something else. The changes will be reflected in a new input session. I get the candidate part, but how often does a user go back to previously typed (now boosted or excluded) phrases, and how much is it really a problem for the candidates to not reflect the boosted/excluded new state?

This PR adds quite some complexity both in the model (the grid) and the "view", and so it'd be helpful to understand whether this is a problem that warrants the cost.

skyeyvapor:
Thanks, that's a fair point.

I originally noticed this while experimenting with allowing an existing phrase in the composing buffer to be selected and excluded with the Delete key. While looking into how exclusion was already implemented, I discovered the `+` / `-` actions in the candidate window, and then noticed this behavior.

The usability issue I had in mind is mainly immediate feedback and consistency: after a user adds, boosts, or excludes a phrase, reopening the candidates for the same reading within the current input session can still show the old state. For example, a phrase that was just excluded can still appear in the candidate list, which can make it look as though the action did not take effect.

That said, I agree with your point about the typical input session being short. I don't have evidence that users frequently revisit the affected reading within the same session, and this PR does add quite a bit of complexity to keep the grid, candidate overrides, and current composition consistent.

PR #909, opened 2026-09-02
feat(input): add Option-key direct input

skyeyvapor:
## Summary

This PR adds an opt-in preference for directly entering letters, numbers, punctuation, and symbols with Option while using McBopomofo.

The preference is disabled by default to preserve existing behavior.

## Motivation

When typing Chinese, users often need to temporarily enter a few half-width letters, numbers, punctuation marks, or symbols, such as identifiers, URLs, commands, or inline code. Switching to a Latin input source and then switching back interrupts the typing flow for these brief inputs.

This option allows users to hold Option and enter the base keyboard-layout output directly while remaining in McBopomofo. It makes mixed Chinese and English input more convenient without introducing a persistent English mode or changing the default behavior.

## Behavior

When the preference is enabled:

- Option with a valid text key directly commits the key's base keyboard-layout output.
- An existing composition is committed before the direct input.
- Option-Shift, Option-Command, and Option-Control combinations retain their existing behavior.
- Spaces, newlines, function-key characters, and other unsupported text are not intercepted.

## Implementation

- Add `optionKeyDirectInputEnabled` to `Preferences`, defaults, snapshots, and diagnostic output.
- Add the preference to the Basic preferences UI with English, Base, and Traditional Chinese localizations.
- Derive `directInputText` with `NSEvent.characters(byApplyingModifiers: [])`.
- Validate the complete translated string as alphanumeric, punctuation, symbol, or non-base Unicode text.
- Use the translated text instead of deriving output from a physical key code.
- Force-commit an existing composition before committing the direct-input text.

## Testing

Added coverage for:

- Disabled preference behavior.
- Valid single-character, multi-scalar, and multi-character text.
- Empty, whitespace-containing, control, and function-key text.
- Keyboard-layout-translated output instead of physical key labels.
- Option dead keys for E, U, I, and N.
- Existing composition commit order.
- Preservation of Option-Shift, Option-Command, and Option-Control combinations.
- Preference defaults and `UserDefaults` persistence.

Focused KeyHandler and Preferences tests pass, and `git diff --check` reports no whitespace errors.

### DIFF
--- a/Source/KeyHandler.mm
+++ b/Source/KeyHandler.mm
@@ -367,6 +367,18 @@ InputMode InputModePlainBopomofo = @"org.openvanilla.inputmethod.McBopomofo.Plai
         return YES;
     }
 
+    if (Preferences.optionKeyDirectInputEnabled && input.isOptionHold &&
+        !input.isShiftHold && !input.isCommandHold && !input.isControlHold) {
+        NSString *directText = input.directInputText;
+        if (input.isValidDirectInputText) {
+            [self handleForceCommitWithStateCallback:stateCallback];
+            InputStateCommitting *committingState = [[InputStateCommitting alloc] initWithPoppedText:directText];
+            stateCallback(committingState);
+            stateCallback([[InputStateEmpty alloc] init]);
+            return YES;
+        }
+    }
+
     // if the inputText is empty, it's a function key combination, we ignore it
     if (!input.inputText.length) {
         return NO;
--- a/Source/PreferencesUI/PreferencesView.swift
+++ b/Source/PreferencesUI/PreferencesView.swift
@@ -499,6 +499,12 @@ private struct BasicPreferencesView: View {
                 .fixedSize()
             }
 
+            PreferenceRow(localized("Option + Key:")) {
+                Toggle(
+                    localized("Input letters, numbers, punctuation, and symbols directly"),
+                    isOn: $preferences.optionKeyDirectInputEnabled)
+            }
+
             PreferenceRow(localized("Shift + Enter Key:")) {
                 Toggle(
                     localized("Trigger associated phrases"), isOn: $preferences.shiftEnterEnabled)
--- a/Source/zh-Hant.lproj/Localizable.strings
+++ b/Source/zh-Hant.lproj/Localizable.strings
@@ -212,6 +212,8 @@
 "Shift + Letter Keys:" = "Shift + 字母鍵：";
 "Input uppercase letters directly" = "直接輸入大寫字母";
 "Input lowercased letters to buffer" = "在輸入緩衝區中輸入小寫字母";
+"Option + Key:" = "Option + 按鍵：";
+"Input letters, numbers, punctuation, and symbols directly" = "直接輸入字母、數字、標點與符號";
 "ESC Key:" = "ESC 鍵：";
 "Beep upon input error" = "輸入錯誤時發出警示聲";
 "Trigger associated phrases" = "使用聯想詞功能";
--- a/McBopomofoTests/KeyHandlerBopomofoTests.swift
+++ b/McBopomofoTests/KeyHandlerBopomofoTests.swift
@@ -22,6 +22,7 @@
 // OTHER DEALINGS IN THE SOFTWARE.
 
 import CandidateUI
+import Carbon.HIToolbox
 import XCTest
 
 @testable import McBopomofo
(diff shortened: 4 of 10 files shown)


Source/KeyHandler.mm
```diff
@@ -367,6 +367,18 @@ - (BOOL)handleInput:(KeyHandlerInput *)input state:(InputState *)inState stateCa
         return YES;
     }
 
+    if (Preferences.optionKeyDirectInputEnabled && input.isOptionHold &&
```
lukhnos, line 370: Although Preferences lookup is fast, but it's not going to be faster than determining the input states. Given `input.isOptionHold` is most often `false`, let's move the first condition to the end of the predicate (that is `input.isOptionHold && ... && Preferences.optionKeyDirectInputEnabled`).

In general, when readability is not affected, let's always try to make the `if` bail early upon most often false conditions.

Source/KeyHandler.mm
```diff
@@ -367,6 +367,18 @@ - (BOOL)handleInput:(KeyHandlerInput *)input state:(InputState *)inState stateCa
...
+        !input.isShiftHold && !input.isCommandHold && !input.isControlHold) {
```
lukhnos, line 371: I wonder if we should take shift into consideration. See my overall comment.

Source/PreferencesUI/PreferencesView.swift
```diff
@@ -499,6 +499,12 @@ private struct BasicPreferencesView: View {
                 .fixedSize()
             }
 
+            PreferenceRow(localized("Option + Key:")) {
```
lukhnos, line 502: This should probably live in `AdvancedPreferencesView` instead.

Also see my overall comment.

Source/PreferencesUI/PreferencesView.swift
```diff
@@ -499,6 +499,12 @@ private struct BasicPreferencesView: View {
...
+                Toggle(
+                    localized("Input letters, numbers, punctuation, and symbols directly"),
```
lukhnos, line 504: The current English l10n string is too long and makes this the only setting that has a two-line description. How about just "Input the key character directly"?

This is also one advantage of moving this to the Advanced view, where you get to explain a bit more if needed (though probably not necessary in this case), see the "Punctuation Symbols" setting there.

Source/zh-Hant.lproj/Localizable.strings
```diff
@@ -212,6 +212,8 @@
 "Shift + Letter Keys:" = "Shift + 字母鍵：";
 "Input uppercase letters directly" = "直接輸入大寫字母";
 "Input lowercased letters to buffer" = "在輸入緩衝區中輸入小寫字母";
+"Option + Key:" = "Option + 按鍵：";
+"Input letters, numbers, punctuation, and symbols directly" = "直接輸入字母、數字、標點與符號";
```
lukhnos, line 216: Does the zh-Hant l10n also result in two-line description?

lukhnos (changes requested):
Overall looking good! Thank you for this useful addition for those who need it.

Regarding usability: `Opt+Shift` is currently not handled. This means you can't really type half-width punctuation symbols with it. When this setting is enabled, `Opt-Shift-/` (so `Opt-?`) gives you the inverted question mark right now, which is how the key combo usually works without the setting. That feels weird if the promise of this PR is to "input letters, numbers, punctuation, and symbols directly." Should we consider supporting `Opt+Shift` too?

Also, and this is my opinion, I suggest this be moved to the Advanced View in the Preferences, as there are already a lot of things the majority of the user cares about, and let's not make it more crowded than necessary. There's also an issue about string length (so UI refinement) that should be addressed.

skyeyvapor:
Thanks for the feedback! After thinking about this more, I’m starting to feel that this may not actually be a good feature to add.

The main issue is that touch-typing behavior is different between Chinese and English input modes. To type the intended English character with this feature enabled, I would need to be consciously aware that I’m currently in the Chinese input method. Otherwise, the same keystroke could produce a symbol instead of the character I expect.

That makes the behavior somewhat context-dependent and less predictable than I originally intended. Given that, I’m leaning toward not adding this feature after all.

lukhnos:
> That makes the behavior somewhat context-dependent and less predictable than I originally intended. Given that, I’m leaning toward not adding this feature after all.

Understood, and thanks again for the PR!

PR #859, opened 2026-07-01
Adds Chinese characters in CJK unified ideographs extension.

zonble:
These characters are selected by

- If it exists in PingFang TC font.
- If its origin is marked as Taiwan in cns11643.gov.tw (全字庫).
- The readings are from 全字庫 as well.

These characters are rarely used by they may exist in names.

PR #911, opened 2026-09-02
Tunes the order of the characters with reading Yi4.

zonble:
We placed some rara characters before popular ones in BPMFBase.txt. In Smart Bopomofo, the rare characters could be placed to the end since we have additional scores for them. However, the traditional Bopomofo uses the original order. The order effects the traditional Bopomofo users.

Fixes for #910.


lwhsu:
Please also check #912 , those characters are added twice, and ㄌㄩˇ has the same issue.

lukhnos:
It seems those are not the only affected entries caused by #859. After the PR got merged, @ChiahongHong made the [comment here](https://github.com/openvanilla/McBopomofo/pull/859#issuecomment-5093192358) re the duplicate entries.

Do we know if @ChiahongHong's findings were the only entries that were problematic? Or should be roll back #859 to be cautious?

Source/Data/BPMFBase.txt
```diff
@@ -19827,6 +19723,110 @@
...
+螠 ㄧˋ yi4 u4 utf8
+袣 ㄧˋ yi4 u4 utf8
+褹 ㄧˋ yi4 u4 utf8
+訲 ㄧˋ yi4 u4 utf8
+譩 ㄧˋ yi4 u4 utf8
+讛 ㄧˋ yi4 u4 utf8
+豛 ㄧˋ yi4 u4 utf8
+貖 ㄧˋ yi4 u4 utf8
+鈠 ㄧˋ yi4 u4 utf8
+阝 ㄧˋ yi4 u4 utf8
```
lukhnos, line 19819: This is not the correct fix, as it only moves the entries, but for example this line is a dup of line 19715.

lukhnos (commented):
Closing this in favor of #912, which is the correct fix.

PR #912, opened 2026-09-02
Remove duplicate entries in the ㄧˋ and ㄌㄩˇ blocks

lwhsu:
Commit f78da56 added new CJK characters.  It added the same 104-character list twice in the ㄧˋ block.  It added the same 13-character list twice in the ㄌㄩˇ block.

Each list has two copies.  One copy sits right after the first entry.  The other copy sits at the end of the block.

The copy in the middle changes the candidate order.  In Plain Bopomofo mode, candidates show in the same order as this file. The middle copy pushed common characters far down the list, for example:

- 義 moved from position 2 to position 106 in the ㄧˋ block.

Remove the middle copy in both blocks so that common characters return to their old position.

This also affects fcitx5-mcbopomofo.  That project copies the compiled data-plain-bpmf.txt from this repo.  It got the same order change after' its last data sync.


lukhnos (approved):
LGTM. Thank you very much for your quick PR.

I have independently verified that this is the correct fix, with 104 `ㄧˋ` and 13 `ㄌㄩˇ` duplicate entries causing the problem.

lukhnos:
@lwhsu We've also just tagged and released fcitx5-mcbopomofo 3.1.1. Thanks again for your help!

PR #915, opened 2026-09-17
feat(input): switch input sources with Shift + letter or Command

skyeyvapor:
## Summary

Add two optional ways to switch from McBopomofo to a selected input source:

- Press Shift + a letter.
- Press either Command key.

Each option has its own target input source. Existing behavior remains the default.

## Motivation

Users who frequently switch between Chinese and English may need to check the current input source before typing, which can interrupt the typing flow.

An uppercase letter often marks the beginning of a stretch of English text. With Shift + letter switching enabled, users can start typing English without first checking whether McBopomofo or the English keyboard is active.

Similarly, a Command shortcut is often followed by English input. Switching on Command press puts the input source into a predictable state before typing continues. Command can also serve as a quick way to switch to English input.

With both options enabled, users can switch between Chinese and English with much less attention to the current input source.

## Behavior

- **Shift + letter:** Switch to the selected input source and return the original event as unhandled, rather than inserting the letter from McBopomofo. Existing candidate, associated-phrase, and marking handlers retain priority.
- **Command:** Switch immediately when either Command key is pressed, including when other modifiers are held. This option is available in Advanced preferences.
- Pending composition is committed before switching.
- The target menus list enabled, selectable, ASCII-capable keyboard layouts. A saved target that becomes unavailable remains visible with an unavailable label.

## Notes

`InputState.SwitchingInputSource` carries an explicit target source ID and reuses the existing empty-state cleanup before switching. Both features use `TISSelectInputSource` within the main input method process.

Testing exposed a lifecycle inconsistency: source selection could succeed without the original controller receiving `deactivateServer`. Returning to McBopomofo could then deliver `setValue` without a new `activateServer` callback. With Vietnamese as the other layout, the number-row 8 key could arrive as `´` instead of producing ㄚ.

To restore the configured keyboard layout in that situation, `setValue` now calls `overrideKeyboard` even when McBopomofo's internal input mode has not changed. Clearing the engine and changing the internal mode remain conditional on an actual mode change. This is a defensive layout restoration, not a claim that the underlying IMK lifecycle issue is resolved.

Returning `false` leaves the letter's subsequent handling to the system/client input pipeline; it does not explicitly redispatch the event to another input method.

## Testing

Three tests in `KeyHandlerBopomofoTests` cover Shift + letter switching:

- `testUppercaseLetterWhenEmpty3`: requests the configured input source, returns the event as unhandled, and emits no letter commit.
- `testUppercaseLetterWhenNotEmpty3`: requests a switch with pending composition and clears the engine.
- `testUppercaseLetterAfterDeletingComposition`: verifies switching after Backspace leaves `EmptyIgnoringPreviousState`.

These tests cover KeyHandler behavior. They do not verify composition delivery to the client, actual system input source selection, or IMK lifecycle callbacks.

Manual development testing covered switching between McBopomofo and Vietnamese and checking keyboard layout restoration after returning to McBopomofo.

Command switching is not covered by automated tests.


lukhnos (commented):
Hi! Thanks for the PR.

This is a complex feature. I've read your Motivation, but could you let us know whether this feature is based on your personal habit and daily usage (so it "scratches an itch"), or was this developed in the hope of solving a problem that someone else had (and if so, who they might be)?

I've spent quite some time with the PR like I did with https://github.com/openvanilla/McBopomofo/pull/909. This PR is actually even more complex as it introduces a very different mental model, its underlying operations have failure modes, and as the PR stands it touches a lot of the Preferences UI, and that in turn has a lot of implications and impact.

It may be better if you can help us understand the "why" and "for whom" of this PR before I comment further. I'm in support of a community-proposed feature that helps people who need it (again, see #909), but I'd also want to make sure that a PR is well understood and done properly, especially if it's on the complex end of the spectrum.

The closest thing McBopomofo has regarding keyboard layout is this: https://github.com/openvanilla/McBopomofo/wiki/使用說明_進階鍵盤配置 ; we treat it as an advanced feature that requirues a user to enable it manually. I just want to mention it because it may serve as a reference point.

skyeyvapor:
Thanks for the thoughtful feedback. This feature did start from my own daily usage. I switch between Chinese and English frequently, and I found that having to keep track of the current input source can interrupt the typing flow.

I don't have usage data to show how common this problem is, so I don't want to overstate it. My sense is that this friction is probably shared by other users who frequently mix Chinese and English, and I think the feature could be useful to a broader group if exposed appropriately.

I also understand your point about the advanced keyboard configuration as a reference. I agree that this kind of behavior is probably better treated as an opt-in advanced feature rather than changing the default mental model for everyone. I'm open to reducing its footprint in Preferences or narrowing the scope if that would make the feature easier to understand and maintain.

PR #917, opened 2026-09-18
fix(braille): handle incomplete indicators

Hokila:
## Summary

Prevent the Braille decoder from crashing when an incomplete number or uppercase indicator is passed in. Incomplete indicators are preserved as text.

## Reproduce steps

1. Open any macOS application that exposes Services.
2. Enter or paste the incomplete Unicode Braille number indicator `⠼` and select it.
3. Choose **Services → Convert Unicode Taiwanese Braille to Text**.
4. Before this fix, McBopomofo crashes with `String index is out of bounds`.

The same issue can be reproduced with `⠁⠠` (a Braille letter followed by an incomplete uppercase indicator), or with ASCII Braille `#` and `a,` using the corresponding ASCII service.

## Root cause

The decoder indexed `readHead + 1` without checking that a following character existed.

## Fix

Add bounds checks before reading the character following number and uppercase indicators, and add Unicode/ASCII regression tests.

## Tests

- `swift test` in `Packages/BopomofoBraille`
- `xcodebuild -project McBopomofo.xcodeproj -scheme McBopomofo -configuration Debug -destination "platform=macOS" test -quiet`


lukhnos (changes requested):
Thanks for the PR. Before we can review this, could you please remove commit 8489a0e? This PR is opened as a bug fix, but that commit has nothing to do with the fix.

Hokila:
I am very sorry; I accidentally included it. I have fixed it now.

lukhnos (approved):
LGTM. Thank you for the bug fix!

